Issue #12, opened 2014-01-05
Drop the use of TestShell and just use phpunit


markstory:
The CakePHP core code in 3.0 has long dropped the use of the TestShell. I think doing the same for the app skeleton has a few benefits:
- Less to learn. Most PHP developers using CLI runners will know how to use phpunit. Not using custom tools makes it easier to pick up CakePHP.
- Simpler test harness. By using more basic tools there is less magic in the way. Also it allows the test runner to not depend on the host framework which is nice.
- Allows CakePHP to be simpler. There is a non-trivial amount of code in CakePHP to support the TestShell and the web runner. Removing this code and just using `phpunit` lets us provide a simpler framework.
### Web runner

Removing the TestShell and underlying libraries will result in the web test runner breaking as well. I think we have an opportunity to either use [an existing web runner solution](https://github.com/NSinopoli/VisualPHPUnit) or build a standalone application that people wanting a web test runner can use.

PR #34, opened 2014-02-28
Remove web test runner, and only use phpunit.

markstory:
This set of changes implements the ideas in #12. I have mixed feelings about removing the `test.php` file.

In the :+1: side for keeping `test.php`
- It is easier for CLI newbies to get testing working.
- It has a nice coverage view per file.

On the :-1: side of keeping `test.php`:
- It is a snowflake. Both in CakePHP and in PHP in general. No other frameworks maintain their own web runner for PHPUnit. I'm considering either contributing to/forking VisualPHPUnit so we can provide a more general use test runner.
- CakePHP has a non-trivial amount of code dedicated to supporting the web-runner. I'd like to see that go away.

I think most CLI newbies are also people who use full featured IDE's. Most good IDE's have a way of integrating with PHPUnit. IDE integration might be more valuable/useful to this group. It certainly is less work once you get it working.

### DIFF
--- a/Test/bootstrap.php
+++ b/Test/bootstrap.php
@@ -1,0 +1,8 @@
+<?php
+/**
+ * Test runner bootstrap.
+ *
+ * Add additional configuration/setup your application needs when running 
+ * unit tests in this file.
+ */
+require dirname(__DIR__) . '/App/Config/bootstrap.php';
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -4,7 +4,7 @@
 	processIsolation="false"
 	stopOnFailure="false"
 	syntaxCheck="false"
-	bootstrap="./App/Config/bootstrap.php"
+	bootstrap="./Test/bootstrap.php"
 	>
 	<php>
 		<ini name="memory_limit" value="-1"/>


burzum:
>  I'm considering either contributing to/forking VisualPHPUnit so we can provide a more general use test runner.

I agree. Actually it already works with CakePHP3. Guess there are things we can do to make it better or something like that?

> CakePHP has a non-trivial amount of code dedicated to supporting the web-runner. I'd like to see that go away.

Yes, remove it! :+1:

dereuromark:
One would need a plugin for a better VisualPHPUnit integration. Right now it is not too feasible to use.

leighmackay:
@markstory  
> It has a nice coverage view per file.

That is what I personally miss. Would it be possible to have another tab on the wonderful DebugKit that just shows test coverage? might be a silly idea.

dakota:
@leighmackay running phpunit with `--coverage-html` gives you an even nicer coverage view :+1:

leighmackay:
@dakota  Thank you. Yes, I run use that, but miss being able to click and see it.

leighmackay:
You guys are right, it's not needed, forget my request. 

:+1:  keeping it out.

PR #183, opened 2014-12-06
Read error handler config instead of consuming in bootstrap.

ADmad:
This allows the config to be reused to register a new handler post
bootstrapping, by only only modifying required options and keeping the rest
same as those set in config/app.php

Use case: Crud plugin's ApiListener needs to set a custom exception render. Since in 3.x only way to change that is register a new error handler you need access to the error config so that it can only change the exception renderer class and use the rest of the settings same as what developer sets in his app config.

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -103,9 +103,9 @@ ini_set('intl.default_locale', 'en_US');
  */
 $isCli = php_sapi_name() === 'cli';
 if ($isCli) {
-    (new ConsoleErrorHandler(Configure::consume('Error')))->register();
+    (new ConsoleErrorHandler(Configure::read('Error')))->register();
 } else {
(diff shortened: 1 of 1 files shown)


josegonzalez:
Any movement on this?

lorenzo:
not so far @josegonzalez... do you want to take this task?

josegonzalez:
No lol I'm just bumping to see if anyone was gonna do anything :P

AD7six:
A low-effort solution might be to just make consumed config retrievable (move it under the config-key "Consumed"?).

dereuromark:
Until we can resolve this, we maybe should just use read() for now?

markstory:
Using read() is a low effort solution that solves some of the issues described here.

ADmad:
Rebased and merged.

PR #204, opened 2015-01-02
Update to 3.0.0-RC1

jadb:


josegonzalez:
It already retrieves the latest RC, so this isn't necessary.

ceeram:
I prefer keeping current require, -dev also pulls in RC1/2

markstory:
Once we hit a stable release I'd really like to change the requirement to `3.0.*`. Tags exist for a reason, and getting people accustomed to using them is a good thing. Also installing from non dev- sources is much faster.

PR #205, opened 2015-01-03
Add intl check in bootstrap.

markstory:
This check should help people understand what they need to do better than 'Missing class Locale' does. Refs cakephp/cakephp#5553

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -33,6 +33,11 @@ require ROOT . DS . 'vendor' . DS . 'autoload.php';
  */
 require CORE_PATH . 'config' . DS . 'bootstrap.php';
 
+// You can remove this if you have confident you have intl installed.
+if (!extension_loaded('intl')) {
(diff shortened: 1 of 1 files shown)


config/bootstrap.php
```diff
@@ -33,6 +33,11 @@
  */
 require CORE_PATH . 'config' . DS . 'bootstrap.php';
 
+// You can remove this if you have confident you have intl installed.
```
dereuromark: "have confidence" or "are confident" :)

dereuromark:
:+1:

antograssiot:
don't this must be before the require autoload.php ?

config/bootstrap.php
markstory: Derp...

markstory:
@antograssiot why? Before the autoloader? The locale class won't be called until it is first used. I will uninstall my intl ext to see if I am dumb.

antograssiot:
@markstory sorry I'm the dumb one, tested it to fast. didn't mean wasted your time

PR #208, opened 2015-01-13
Add stub for plugins configuration file.

markstory:
This config file stub will be used with cakephp/cakephp#5641

I'm still working on the plugin-installers change. I'm hoping to have it done soon.

### DIFF
--- a/config/plugins.php
+++ b/config/plugins.php
@@ -1,0 +1,23 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.0.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+
+/**
+ * This file contains a list of plugins and their install paths.
+ *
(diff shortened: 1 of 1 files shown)


config/plugins.php
```diff
@@ -0,0 +1,23 @@
...
+ * @since         3.0.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+
+/**
+ * This file contains a list of plugins and their install paths.
+ *
+ * This is primarily used for identifying plugins in /vendor
+ */
+return [
```
ADmad: `PhpConfig` engine expects the file to have [$config](https://github.com/cakephp/cakephp/blob/3.0/src/Core/Configure/Engine/PhpConfig.php#L71) variable.

lorenzo: Should we add support for this in PHPConfig? I feel like returning is more natural than using a variable.

ADmad: I agree, returning is more natural. We could get rid of `$config` all together?  Shouldn't be a problem to check for it and show an error to developers so that they can update their existing files.

markstory: Bah I suck. I can update this for now.

ADmad: You shouldn't have. Just merge cakephp/cakephp#5658 instead :)

config/plugins.php
```diff
@@ -0,0 +1,23 @@
...
+ * @since         3.0.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+
+/**
+ * This file contains a list of plugins and their install paths.
+ *
+ * This is primarily used for identifying plugins in /vendor
+ */
+$config = [
```
ADmad: Now that my PhpConfig engine PR is merged can we just change this to
```php
return [
    'plugins' => []
]
```

AD7six: app.default.php needs changing too for that.

AD7six:
Merging, I'll change the config files to return in a separate commit.

PR #210, opened 2015-01-14
Update app.php to return array instead of setting $config var.

ADmad:
Refs cakephp/cakephp#5658


AD7six:
closing in favor of #211 (er.. oops wouldn't have done that had I seen this :trollface:).

PR #213, opened 2015-01-16
Rename config/Schema to config/schema.

ADmad:
Given that the folder doesn't contain any classes using lowercase makes seems better.


dereuromark:
:+1:

PR #214, opened 2015-01-17
Ignore config/plugins.php

jadb:
This file gets overwritten by composer on every deploy anyways...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /vendor/*
 /config/app.php
+/config/plugins.php
 /tmp/*
 /logs/*


ADmad:
:+1:

AD7six:
What was the purpose of this..?

You can't ignore a [tracked file](https://github.com/cakephp/app/tree/master/config) - it doesn't do anything if you do.

ADmad:
@AD7six Isn't this the same was what we have done for `tmp` folder?  `/tmp/*` is ignored yet it's subfoldesr with `empty` files are tracked.

AD7six:
[No](https://gist.github.com/AD7six/76f3ed193f1549309ac1). Additional _untracked_ files in the tmp folder are ignored.

jadb:
I think AD7six is right - I have overlooked that. Did not notice it was already added to the repo. I think it should not be tracked at all, what do guys you think?

lorenzo:
@AD7six you don't get a git repo after creating an app with composer. As soon as you do `git init && git add .` that file will be excluded

AD7six:
@lorenzo only if you don't do this:

```
ww-data @ dev [ /tmp ]
-> composer create-project cakephp/app


                                                             
  [InvalidArgumentException]                                 
  Could not find package cakephp/app with stability stable.  
                                                             


create-project [-s|--stability="..."] [--prefer-source] [--prefer-dist] [--repository-url="..."] [--dev] [--no-dev] [--no-plugins] [--no-custom-installers] [--no-scripts] [--no-progress] [--keep-vcs] [--no-install] [--ignore-platform-reqs] [package] [directory] [version]


www-data @ dev [ /tmp ]
-> composer create-project cakephp/app -s dev
...
Do you want to remove the existing VCS (.git, .svn..) history? [Y,n]? n ##### <---
www-data @ dev [ /tmp ]
-> git status
fatal: Not a git repository (or any of the parent directories): .git
www-data @ dev [ /tmp ]
-> cd app/
www-data @ dev [ /tmp/app ] (master *%=)
-> git status
# On branch master
# Changes not staged for commit:
#   (use "git add <file>..." to update what will be committed)
#   (use "git checkout -- <file>..." to discard changes in working directory)
#
#	modified:   bin/cake.bat
#	modified:   config/plugins.php
#
# Untracked files:
#   (use "git add <file>..." to include in what will be committed)
#
#	composer.lock
no changes added to commit (use "git add" and/or "git commit -a")
www-data @ dev [ /tmp/app ] (master *%=)
-> 
```

I think it's a bit odd to ignore a file that is tracked - on the assumption that users will create a new git repository for their project.

Isn't `plugins.php` a required file:

```
www-data @ dev [ /tmp/app ] (master *%=)
-> rm config/plugins.php 
www-data @ dev [ /tmp/app ] (master *%=)
-> bin/cake 
Exception: Plugin Migrations could not be found. in [/tmp/app/vendor/cakephp/cakephp/src/Core/Plugin.php, line 149]
www-data @ dev [ /tmp/app ] (master *%=)
````

?

It does not seem like a good idea to me to make it harder to commit a file an application depends on (assuming users don't start with the existing git history).

Is the bottom-line reason to make this change to address the file being dirty when creating a new project? Travis noise? Something else? IMO it should exist in the app's git repo in the state that permits it to work.

lorenzo:
The reason is that is not actually needed in the repo. I'm wondering if instead of putting it in config we should have it in `vendor` side by side with `autoload.php`

dereuromark:
@lorenzo That makes a lot more sense indeed.

markstory:
The config file is not required in the repo. The core libraries will work even if the file is missing from what I remember.

AD7six:
@markstory unless you try to use any plugin.

I've modified the plugin installer to dump the config file when dump-autoload is called: cakephp/plugin-installer:16

That also opens the door to moving the plugins config file into vendors/x I think.

markstory:
Why can we not move the `plugins.php` file into vendors/ as things are? Other than having to setup a specific PhpConfig instance - which we should probably do anyways I can't think of a reason the existing file couldn't be put in vendor/ instead of config/.

PR #216, opened 2015-01-18
Use the autoload-dump hook

AD7six:
Rather than incrementally add to the plugins file whenever you add or
remove a plugin; generate a complete plugins file every time the
autoloader is dumped (which is all of when called manually, when adding
a vendor, when removing a vendor etc.)

This removes the need for the config/plugins.php file to exist in the
repository (and in anticipation of it being permanently moved to the
vendors folder, the file and the ignore are removed)

related cakephp/plugin-installer#16


dereuromark:
Cool :)

markstory:
The downside to this is everyone needs to sync up their application's composer.json file.

ADmad:
:+1: to the hook. :-1: to adding the `postAutoloadDump()` method in plugin installer. It could just be in appropriate class of core or app.

AD7six:
If it's in the app repo - we kind of loose control of being able to update it.

if it's in cakephp/cakephp - that means adding [a lot of composer dependencies](https://github.com/cakephp/plugin-installer/blob/dump-autoload-hook/src/Installer/PluginInstaller.php#L5-L11) that don't currently exist. Though it would make sense to me to put the logic responsible for generating a pseudo-config file in the same repo as the one that needs to read it.

ADmad:
The hooks methods only explicitly only require `Composer\Script\Event;` and `\Composer\Package\PackageInterface` (if we typehint `$packages` arg of `determinePlugins`).

ADmad:
We could keep the hook methods in a separate repo but imo then shouldn't be in a class which implements `Composer/Installer/InstallerInterface.php` because that's totally irrelevant.

Also the current scenario is you need to require this plugin installer in plugin but should not specify `type: "cakephp-plugin"` else it will run the instance methods and show warnings. This can be quite confusing to users.

markstory:
I don't get what the upside of generating the config file with the autoloader hook is. The current implementation works, and the main issue I can see with the existing code ia where the config file is saved.

AD7six:
@markstory The problem I see is we're now ignoring `plugins.php` - so if you have an otherwise complete install, but don't have this file how can it be recreated?

The existing solution only does anything on `composer require`/`composer remove` - and only for the one vendor-plugin being actioned. We don't have a means to generate `plugins.php` without nuking the vendors dir and reinstalling everything or writing this file by hand.

I feel that's quite a fragile situation to put users in hence it'd be a good idea to provide them a way to create it in it's entirety -> use the post-autoload-dump hook.

composer.json
```diff
@@ -31,7 +31,8 @@
         }
     },
     "scripts": {
-        "post-install-cmd": "App\\Console\\Installer::postInstall"
+        "post-install-cmd": "App\\Console\\Installer::postInstall",
+        "post-autoload-dump": "Cake\\Composer\\Installer\\PluginInstaller::postAutoloadDump"
```
lorenzo: This would basically mean that this repo depends on plugin-installer, no?

mtkocak:
how should I change the composer.json file?

lorenzo:
@mtkocak look at the composer.json file in this repo

davidyell:
Yeah, pointing people to this PR from composer doesn't really detail the actions you need to take.

AD7six:
@davidyell I've added a comment to the issue description - does that help?

It's a bit confusing for me how linking users to a diff indicating the change they need to make is insufficient.

davidyell:
Yeah @AD7six true, but usually with a change like this it would be a book page. I'm a developer and I didn't even think to look at the actual commit!

AD7six:
@davidyell it's a transient problem, only affecting users who have 3.x apps created before 23rd jan.

It was hoped that linking to the PR indicating the changes they needed to make would be sufficient - and being a transient problem, wasn't worthy of putting in the book (since it's already obsolete info).

davidyell:
Ah ok, I see. Sounds fair :smile:

chrmina:
Your composer.json file should look like this:
```json
{
	"name": "cakephp/upgrade",
	"description": "CakePHP upgrade tool",
	"homepage": "http://cakephp.org",
	"type": "project",
	"license": "MIT",
	"require": {
		"cakephp/cakephp": "~3.0",
		"cakephp/app": "~3.0",
		"phpunit/phpunit": "*",
		"cakephp/plugin-installer": "dev-master"
	},
	"autoload": {
		"psr-4": {
			"Cake\\Upgrade\\": "src",
			"Cake\\Upgrade\\Test\\": "tests"
		}
	},
	"minimum-stability" : "dev",
	"prefer-stable": true,
	"support": {
		"source": "https://github.com/cakephp/upgrade"
	},
    "scripts": {
        "post-autoload-dump": "Cake\\Composer\\Installer\\PluginInstaller::postAutoloadDump"
    }	
}
```

PR #217, opened 2015-01-18
Remove unnecessary `else`

alecho:


lorenzo:
So after asking if you don't want to set the permissions it will set the permissions again?

alecho:
Ah your right. I read those lines a couple times and I thought I was right, hence the PR. I'll close.

jadb:
@lorenzo that's about right. It will set the permissions if you choose not to (as you said) and will double-set them if you choose to set them in the first place (interactive mode). 

That been said, I don't think that's the intended behaviour. Is it?

jadb:
hmm - you guys were quicker :+1:

PR #221, opened 2015-02-09
Ignore /plugins folder

bravo-kernel:
I just noticed all plugins appear in my git status which does not feel right. This PR ignores the /plugins directory and sorts entries in the .file.

Just close if not desired.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-/vendor/*
 /config/app.php
-/tmp/*
 /logs/*
+/plugins/*
+/tmp/*
+/vendor/*


dereuromark:
I don't think this is necessary anymore with the recent changes of plugins going into vendor now.
This change would make baked non-composer plugins un-commitable - but they should be (and was the reason for the vendor move of composered ones in the first place).

AD7six:
Only non-vendor plugins should go in that folder, this was suggested in the past when both vendor and app plugins were stored in the same place, but that's no longer the case.

Duplicate of https://github.com/cakephp/app/pull/206

bravo-kernel:
Thanks for clarifying, sorry for duplicating.

bravo-kernel:
Just for the records; emptying my plugins folder and re-running composer update did indeed install everything in /vendor. Except for DebugKit but that's probably me.

PR #222, opened 2015-02-13
add error and success Flash in default layout

cake17:
As we have elements `error` and `success`, maybe we can add them in the default layout ?


ADmad:
The first param of FlashHelper::render() is the session key not element name. Irrespective of the error type and element use by default the `flash` key will be used.

ADmad:
I added comment on your plugin's issue.

PR #225, opened 2015-03-03
Add description to non-debug home exception

bravo-kernel:
Simply not displaying the home page when `debug` is turned off seems counter intuitive (to new users). 

Today, a user was (exactly like me some time ago) investigating routes and stuff after disabling debug and getting a blank page. 

I understand the reason behind protecting users from going live with a "forgotten" home page but think new users should be protected from confusion by either:

- pointed to the cause using some sort of message in the Exception
- removing the condition alltogether (and expect users going live to be well educated enough to disable/remove home if unneeded)

### DIFF
--- a/src/Template/Pages/home.ctp
+++ b/src/Template/Pages/home.ctp
@@ -20,10 +20,6 @@ use Cake\Network\Exception\NotFoundException;
 
 $this->layout = false;
 
-if (!Configure::read('debug')):
-    throw new NotFoundException();
-endif;
-
 $cakeDescription = 'CakePHP: the rapid development php framework';
 ?>
 <!DOCTYPE html>


ceeram:
The default home.ctp traffic lights page is not meant to be used in production, perhaps instead add a message to exception, like with the old test.php file:

```
 throw new NotFoundException('Debug setting does not allow access to this url.');
```

AD7six:
I definitely do not want this page potentially-accessible on all production sites.

> and getting a blank page.

A blank page? it should be a standard 4oh4:

screen shot 2015-03-03 at 13 17 20

> removing the condition alltogether (and expect users going live to be well educated enough to disable/remove home if unneeded)

I think it'd be irresponsible to make accessible a page advertising information about the install - as such :-1: to the PR in it's current form (and to any change which modifies the publicly accessible response for that view file).

bravo-kernel:
Great, that option had my preference as well. It would be nice if the message could be a bit more descriptive towards the cause like:

    Debug setting in home.ctp does not allow access to this url.

bravo-kernel:
@AD7six you are correct, it should have said 404 and.. glad to see you're not in favor of dropping it altogether as well. If the text I suggested is acceptable I will update the PR.

ADmad:
There is no need for any change. The change you suggest provides internal info about the app to the end user.

bravo-kernel:
The fingerprint argument is valid but displaying a message seems like a valid compromise to aid new users. I'm not sure about real life but IMO live apps should not even be having home.ctp anymore or at least have it altered.

ADmad:
> I'm not sure about real life but IMO live apps should not even be having home.ctp anymore or at least have it altered

Exactly. But in case some does that mistake, the particular check shows an error page without divulging any internal info.

markstory:
I am :-1: on this for the reasons listed above.

bravo-kernel:
Points taken and all valid but... not giving up yet.

Would adding a notice about home not displaying in non-debug mode here be an option?

cake-comment

AD7six:
@bravo-kernel If that's helpful to new users - I don't object to it.

bravo-kernel:
I really do think so, it's the first point of contact and probably what most new users start playing around with. 

If no one objects... can this PR be re-opened so I can revert and apply or should I create a new one?

ADmad:
@bravo-kernel Creating a new one will be easier.

PR #226, opened 2015-03-15
Use `psysh` for `cake console` instead of `boris`

antograssiot:
so even windows users can enjoy this feature without using a VM.
http://psysh.org/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "cakephp/plugin-installer": "*"
     },
     "require-dev": {
-        "d11wtq/boris": "1.0.*",
+        "psy/psysh": "@stable",
         "cakephp/debug_kit": "3.0.*-dev",
         "cakephp/bake": "dev-master"
     },
--- a/src/Shell/ConsoleShell.php
+++ b/src/Shell/ConsoleShell.php
@@ -32,7 +32,6 @@ class ConsoleShell extends Shell
      */
     public function main()
     {
-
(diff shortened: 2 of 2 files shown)


antograssiot:
I took care of th CS error

kamui545:
I was thinking about the same few days ago, I think it's a good move :+1:

markstory:
Looks like a big improvement to me.

lorenzo:
This looks like a big improvement to me as well. We may want to register our own debugger to psysh at some point. I'm going to merge this so we can iterate over it :)

Great jon @antograssiot !

dereuromark:
It will need a few extensions to work flawless it seems:
```
psy/psysh suggests installing ext-pcntl (Enabling the PCNTL extension makes PsyS
H a lot happier :))
psy/psysh suggests installing ext-posix (If you have PCNTL, you'll want the POSI
X extension as well.)
psy/psysh suggests installing ext-readline (Enables support for arrow-key histor
y navigation, and showing and manipulating command history.)
psy/psysh suggests installing ext-pdo-sqlite (The doc command requires SQLite to
 work.)
```
But looks good and I will give it a spin as soon as possible :)

ceeram:
@dereuromark not very different from boris: https://github.com/borisrepl/boris/blob/master/composer.json

ceeram:
On boris they are hard requirements instead of suggests

PR #228, opened 2015-03-24
Add PHP version check in cake.php

rchavik:
This presents a friendlier error message for people accidentally
running it with 5.3.  Since this check is in bin/cake.php, it should
not affect normal web requests.

Related: cakephp/cakephp#6157

Note: No `3.1` branch yet, so this is still targeted on `master`.

### DIFF
--- a/bin/cake.php
+++ b/bin/cake.php
@@ -15,6 +15,19 @@
  * @since         2.0.0
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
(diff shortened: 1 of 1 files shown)


antograssiot:
:+1: Hopefully that will save some users inputs

jadb:
Hmm, maybe I am missing something but doesn't composer do version checks? And isn't cake3 only `composer install`-supported?

PR #231, opened 2015-04-02
Calling parent initialize on controller

jrbasso:
The parent method (Cake's Controller) is empty, but is a good practice to call it in case the parent starts to implement something.

Also, the doc examples also include the parent call, so give more consistency.


ADmad:
:+1:

antograssiot:
:+1:

Issue #237, opened 2015-04-22
Exclude 'empty' files in .gitignore


elboletaire:
I've tried to add some exclusions to .gitignore so the first time you do a `git add .` the `tmp` and `logs` folder are also added.

I was creating a PR with these changes but I'm not able to add the exclusions for the `tmp` folder. It seems that git ignores its subfolders.

Taking the current `.gitignore` file:

    /vendor/*
    /config/app.php
    /tmp/*
    /logs/*

The `empty` file in the `logs` folder can be excluded like so:

    /vendor/*
    /config/app.php
    /tmp/*
    /logs/*
    !/logs/empty

But with the `tmp` folder, I've tried (without success) all these convinations:

    !/tmp/**/empty
    !/tmp/*/*/empty
    !/**/empty # try to exclude every empty file found (does not work with more than one level)
    !empty

As said, none of them work (at least not in my git environment..).

I think that it would be great to find a way to exclude these files as I'm sure I'm not the only one who needs to manually add these folders after the first clone (because `composer install` crashes when tries to apply its permissions).

If we cannot add the `.gitignore` exclusions maybe the folder creation could be managed by the composer post-install script.

markstory:
Couldn't we explicitly list all the empty files that come with the app skeleton out of the box?

bravo-kernel:
Could somebody please confirm explicitly listing all empty files is the solution so I can create the PR?

elboletaire:
Sorry I lost track of this. Glad to see it has been solved.

bravo-kernel:
Hmmm, just noticed that the explicit exclusions in #238 only work for the ``/logs`` directory and not for ``/tmp`` (most likely caused by the subdirectories as described by @elboletaire).

So far I have had no luck in finding the right syntax.

jadb:
Hm, does that mean we should revert #238?

bravo-kernel:
I remember having to do something similar once and it turns out this was the (only) way to do it:
```
/vendor/*
/config/app.php

/logs/*
!/logs/empty

/tmp/*
!/tmp/cache
/tmp/cache/*

!/tmp/cache/models
/tmp/cache/models/*
!/tmp/cache/models/empty

!/tmp/cache/persistent
/tmp/cache/persistent/*
!/tmp/cache/persistent/empty

!/tmp/cache/views
/tmp/cache/views/*
!/tmp/cache/views/empty

!/tmp/sessions
/tmp/sessions/*
!/tmp/sessions/empty

[...]

PR #238, opened 2015-04-23
No longer ignore temp files

bravo-kernel:
As discussed here https://github.com/cakephp/app/issues/237

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,11 @@
 /config/app.php
 /tmp/*
 /logs/*
+
+
+!/logs/empty
+!/tmp/cache/models/empty
+!/tmp/cache/persistent/empty
+!/tmp/cache/views/empty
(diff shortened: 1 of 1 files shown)


jadb:
If we are to ignore those files (which would effectively have them added to any application's repository on the first `git add .`), then why not also include them in the app skeleton instead of creating them in a post-install hook?

markstory:
They were removed by @AD7six a while ago, I don't remember why but remember he had good reasons for omitting them. This seems like a reasonable compromise between not having the files and having them to me.

jadb:
I agree with the `.gitignore` approach, I am just wondering why not also just include the folders `log` and `tmp` in the app skeleton and ditch that part of the post-install hook. Guess I'll have to wait for @AD7six to see this and reply back.

PR #240, opened 2015-04-28
Enable parsing locale dates by default 

dereuromark:
to match auto localization of Time() when used as string in form inputs.
This is relevant for datetime form fields as strings, for dropdowns it is currently working fine out of the box.

Resolves https://github.com/cakephp/cakephp/pull/6351


markstory:
Seems reasonable to me.

CauanCabral:
:+1:

PR #243, opened 2015-05-07
Update debugkit version to get ajax requests display support

ADmad:


lorenzo:
I don't think it is wise to update cakephp at this moment :P

ADmad:
Oh f***, I need more coffee.

ceeram:
:coffee:

PR #244, opened 2015-05-07
Turn on serialization by default for JsonView/XmlView.

ADmad:
Refs cakephp/cakephp#6502

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -49,7 +49,7 @@ class AppController extends Controller
      */
     public function beforeRender(Event $event)
     {
-        if (in_array($this->viewClass, ['Json', 'Xml'])) {
+        if (in_array($this->response->type(), ['application/json', 'application/xml'])) {
             $this->set('_serialize', true);
         }
     }
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -49,7 +49,7 @@ class AppController extends Controller
      */
     public function beforeRender(Event $event)
     {
-        if ($this->viewClass === 'Json') {
+        if (in_array($this->viewClass, ['Json', 'Xml'])) {
             $this->set('_serialize', true);
         }
     }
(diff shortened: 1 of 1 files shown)


src/Controller/AppController.php
```diff
@@ -39,4 +40,17 @@ public function initialize()
...
+     * Before render callback.
+     *
+     * @param \Cake\Event\Event $event The beforeRender event.
+     * @return void
+     */
+    public function beforeRender(Event $event)
+    {
+        if ($this->viewClass === 'JsonView') {
+            $this->set('_serialize', true);
+        }
```
markstory: What about xmlview?

josegonzalez: Also, what if the fully-qualified namespace is used?

ADmad: > What about xmlview?

The check should be updated when same feature is implemented for XmlView too :)

> Also, what if the fully-qualified namespace is used?

RequestHandler sets the viewClass to `JsonView` so I added check for that. If someone is manually setting the fully namespaced class name or a custom class they should update this check accordingly. In general the app skeleton is meant to be customized as per ones needs.

ADmad:
Turned on serialization for XmlView too. 

Instead of checking `viewClass` how about checking for response type instead?

src/Controller/AppController.php
```diff
@@ -39,4 +40,17 @@ public function initialize()
         parent::initialize();
         $this->loadComponent('Flash');
```
lorenzo: Should we add RequestHandler ?

dereuromark: Makes sense.

ADmad: Yup,  will do.

ADmad:
Loaded `RequestHandler`.

markstory:
This looks good :+1:

lorenzo:
@ADmad I just merged this, but the second after I thought a problem with this. This will override any `_serialize` already set from the controller action :(

ADmad:
@lorenzo Fixing it.

PR #251, opened 2015-05-30
New design added together with new bake template

mtkocak:
muted scheme is used

### DIFF
--- a/webroot/css/cake.css
+++ b/webroot/css/cake.css
@@ -399,7 +399,7 @@ pre {
 }
 
 .side-nav li.heading {
-  color: #373737;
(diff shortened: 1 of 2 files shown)


mtkocak:
https://github.com/cakephp/cakephp/issues/6679#issuecomment-107060805

webroot/css/cake.css
```diff
@@ -382,3 +382,175 @@ pre {
   color: red;
   margin-right: 9px;
 }
+
+/* Midori Universal */
```
AD7six: This comment shouldn't be here =)

PR #254, opened 2015-06-06
Updated doc blocks to remove reference to bake and replaced with some…

garethellis36:
…thing more relevant

### DIFF
--- a/bin/cake.bat
+++ b/bin/cake.bat
@@ -1,6 +1,6 @@
 ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
 ::
-:: Bake is a shell script for running CakePHP bake script
+:: Windows batch script for invoking CakePHP shell commands
 ::
 :: CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
 :: Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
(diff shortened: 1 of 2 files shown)


bin/cake.bat
```diff
@@ -1,30 +1,30 @@
...
-SET lib=%~dp0
-
-php "%lib%cake.php" %*
-
-echo.
-
-exit /B %ERRORLEVEL%
+::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+::
+:: Windows batch script for invoking CakePHP shell commands
```
markstory: Looks like newlines may have changed. This file needs to have crlf endings as it is for windows.

garethellis36:
Ah yes. I'll sort that.

bin/cake.bat
garethellis36: Slightly confused as when I open cake.bat in Sublime/PhpStorm, it's showing as CRLF already.

markstory: Odd, the current file shows as CRLF for me in vim as well.

garethellis36: I checked this out on my windows laptop and it shows as CRLF on there (and works fine). Could it be a git setting on my Mac? I could start over and do the bash script on my Mac and the .bat file on Windows to see if it helps :-)

markstory: That might help. The line endings could be changed by mac git/editor for sure.

Issue #253, opened 2015-06-02
There is EmailTransport and Email in app.php


lorenzo:
One of them need to be removed

ADmad:
So nest transport config under `Email` ?

ADmad:
I like having it separate since config for both is independent of each other.

markstory:
Why does one of them need to be removed?

lorenzo:
Or I may not be understanding what the difference is?

markstory:
Email is for configuring the email object defaults.  While EmailTransport is for configuring SMTP server settings, or mailchimp/mandrill credentials.

ADmad:
I don't think anything needs to be done here, both those configs are needed. :)

lorenzo:
should we add a dock block explaining what they are? It wasn't obvious to me at all

markstory:
I can add something to the docs.

ADmad:
Hmm.. the docblock above these configs does have sections "Configuring transports" and "Configuring delivery profiles". Perhaps the section "Configuring delivery profiles" should be moved directly above the config `Email` ?

lorenzo:
I think that would help

markstory:
Closing as I split the docs up as discussed.

PR #256, opened 2015-06-08
Split comments for transport and profile up.

markstory:
Hopefully with separate comments the two config keys don't seem as redundant.

Refs #253

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -150,15 +150,11 @@ return [
     /**
      * Email configuration.
      *
-     * You can configure email transports and email delivery profiles here.
-     *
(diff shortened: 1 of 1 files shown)


config/app.default.php
```diff
@@ -192,6 +181,16 @@
...
 
+    /**
+     * Email delivery profiles
+     *
+     * Delivery profiles allow you to predefine various properties about email
+     * messages from your application and give the settings a name. This saves
+     * duplication across your application and makes maintenance and development
+     * easier. Each profile accepts a number of keys. See `Cake\Network\Email\Email`
+     * for more information.
+     *
```
bcrowe: Extra line?

PR #259, opened 2015-06-08
Updated docblocks in shell scripts

garethellis36:
Diff still suggests line endings have changed but the .bat file is definitely still CRLF (edited on Windows laptop in notepad++ with line-endings visible and commited via Git with global autocrlf=true) \_(ツ)_/


markstory:
I can take a look at the bat file after merging.

PR #260, opened 2015-06-11
Style tweaks

markstory:
* Don't use ID selectors.
* Make sidebar equal height to content region. On long pages they were not equal.
* Add pagination arrows. They might be backwards though, I can never remember which way they are supposed to go.

Related to cakephp/bake#118

### DIFF
--- a/webroot/css/cake.css
+++ b/webroot/css/cake.css
@@ -91,7 +91,7 @@ ul.pagination li a {
 }
 
 ul.pagination li.active a {
-  background: none repeat scroll 0% 0% #DCE47E;
+  background-color: #DCE47E;
   color: #FFF;
   font-weight: bold;
   cursor: default;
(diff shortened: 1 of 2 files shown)


webroot/css/cake.css
```diff
@@ -521,34 +528,47 @@ div.message:before {
...
-#content h3 {
+/* Use 'one true layout' methods to get equal height columns */
+.container {
+  overflow: hidden;
+  min-height: 92%; /* full height almost always */
+}
+
+/* Force equal height by overflowing */
+.content,
+#actions-sidebar {
```
ADmad: I see an id instead of class :)

markstory: I only removed `#content` in this pass.

lorenzo:
Good job!

mtkocak:
did you test the bake generator?

markstory:
@mtkocak I did, is there a problem?

mtkocak:
No that's perfect.

2015-06-11 18:56 GMT+02:00 Mark Story <notifications@github.com>:

PR #264, opened 2015-06-28
Allow ConsoleShell to be --quiet and --verbose

### LINKED ISSUE #263: ConsoleShell: adding a -q / --quiet option?
Doing something like this is currently extremely difficult:

```shell
$ echo 'use Cake\Datasource\ConnectionManager; ConnectionManager::config("default");'
   | bin/cake Console

Welcome to CakePHP v3.0.7 Console
---------------------------------------------------------------
App : src
Path: /var/www/src/
---------------------------------------------------------------
You can exit with `CTRL-C` or `exit`

Psy Shell v0.4.4 (PHP 5.6.10-1+deb.sury.org~precise+1 — cli) by Justin Hileman
=> [
       "className"        => "Cake\\Database\\Connection",
       "driver"           => "Cake\\Database\\Driver\\Mysql",
       "persistent"       => false,
       "host"             => "localhost",
       "username"         => "vagrant",
       "password"         => "vagrant",
       "database"         => "vagrant",
       "encoding"         => "utf8",
       "timezone"         => "UTC",
       "cacheMetadata"    => true,
       "quoteIdentifiers" => false,
       "name"             => "default"
   ]

Exit:  Ctrl+D

```

All of that banner stuff, the `=>` and the `Exit: Ctrl-D` are all seriously in the way if I wanted to consume this output from somewhere else. What I'd really want out of a call like that is this:

```shell
$ echo 'use Cake\Datasource\ConnectionManager; var_export(ConnectionManager::config("default"));'
   | bin/cake Console --quiet
array (
  'className' => 'Cake\\Database\\Connection',
  'driver' => 'Cake\\Database\\Driver\\Mysql',
  'persistent' => false,
  'host' => 'localhost',
  'username' => 'vagrant',
  'password' => 'vagrant',
  'database' => 'vagrant',
  'encoding' => 'utf8',
  'timezone' => 'UTC',
  'cacheMetadata' => true,
  'quoteIdentifiers' => false,
  'name' => 'default',
)
```

Suppressing the banner is easy, but I have not investigated what it might take to suppress Boris's REPL output and **only** print what the provided commands might produce.

I could even see a case being made for `--quiet` being enabled by default when accepting redirected input like this.

beporter:
Removes the second param to `ConsoleOptionParser::__construct()` which was disabling the default "quiet" and "verbose" options.

Fixes #263

### DIFF
--- a/src/Shell/ConsoleShell.php
+++ b/src/Shell/ConsoleShell.php
@@ -65,7 +65,7 @@ class ConsoleShell extends Shell
      */
     public function getOptionParser()
     {
-        $parser = new ConsoleOptionParser('console', false);
+        $parser = new ConsoleOptionParser('console');
         $parser->description(
             'This shell provides a REPL that you can use to interact ' .
             'with your application in an interactive fashion. You can use ' .


markstory:
This will only effect the CakePHP bits, but it is worth having.

PR #265, opened 2015-07-09
remove debug config check in home.ctp and run test against php7

antograssiot:
This check is useless since debug = false will throw an exception https://github.com/cakephp/app/blob/master/src/Template/Pages/home.ctp#L23-L25

and run test against PHP7.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,7 @@ language: php
 sudo: false
 
 php:
-  - 5.4
-  - 5.5
-  - 5.6
+  - 7.0
 
 before_script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,6 +6,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
```
ADmad: I would remove all other php versions beside 7.0. We are only running the phpcs build so running it on multiple php versions is pointless.

antograssiot:
@ADmad sure I've updated it, I thought there were some tests

PR #269, opened 2015-07-30
Revert "I18n cache"

markstory:
We don't need a separate cache in 3.x as the _cake_core_ config is only used for i18n now.

Reverts cakephp/app#268

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -87,8 +87,9 @@ return [
         ],
 
         /**
-         * Configure the cache used for general framework caching. Path information
-         * and object listings are stored with this configuration.
+         * Configure the cache used for general framework caching. Path information,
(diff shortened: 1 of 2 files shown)


config/app.default.php
```diff
@@ -87,8 +87,9 @@
         ],
 
         /**
-         * Configure the cache used for general framework caching. Path information
-         * and object listings are stored with this configuration.
+         * Configure the cache used for general framework caching. Path information,
+         * object listings, and translation cache files are stored with this
```
ADmad: This statement should be updated. No path info or object listings are cached.

markstory: Done in 32e94a8344253e9a4e56bcf98a460266b1f59228

PR #273, opened 2015-08-05
Declare log option

ionas:
So that developers know just by looking at it, that the Datasource's file logging can be turned on.


lorenzo:
Thanks

PR #277, opened 2015-09-01
replace $layout by ->viewBuilder()->layout()

cake17:


ADmad:
Template files are within the View class context itself so they don't need to use view builder. View instance is already constructed.

Issue #275, opened 2015-08-14
3.1 Fash Message Styles


ionas:
Did set cakephp, bake to 3.1 and manually imported 3.1 app skeleton css files.
https://gist.github.com/ionas/7505988a068f208e1373
I have following issue with Flashes not rendering like they did in 3.0 (full width above main content):
<img width="1083" alt="screen shot 2015-08-12 at 13 30 42" src="https://cloud.githubusercontent.com/assets/106845/9223009/b61fc924-40f6-11e5-9cec-9e20c1eac819.png">
Is that an intentional change?

And: Is there a way to enable the beautiful colorful scheme (at least the basic grid background colors and side menu looks a lot more beautiful)?

Refs:
https://github.com/cakephp/cakephp/issues/6679
https://github.com/cakephp/app/pull/251
https://github.com/cakephp/bake/pull/104

ionas:
Can someone confirm this?
Refs: @mtkocak

markstory:
I'm pretty sure its an intentional change. I'm guessing not one you like?

ionas:
Oh I don't have to like it, I just wanted to know if its in there. If I don't like it I can prep PRs (smaller PRs, I learned that :p) I just wanted to know if it was intentional and if you could duplicate the behavior on your side.

antograssiot:
@ionas I think it was intentional, it is working the same way for me.
Regarding the scrollbar for debug kit, it has nothing to do with the logo, it is just to give you access to the variable panel. It's not so nice but when the browser is not large enough, it was the easier way to still access all the panels.

mtkocak:
That's intentional.

ionas:
Closing then.

ionas:
The top right menu bar is inaccessible when the flash renders "ontop" of it and does not even feature a :hover {display: none};

markstory:
Adding a hover style or a small javascript based close button should be simple enough.

PR #280, opened 2015-09-07
Cleanup & improve the CSS for flash messages.

markstory:
* Remove duplicate CSS for flash messages.
* Add some basic JS to hide flash messages on click.
* Use a basic height rollup.

flash-messages

Refs #275


antograssiot:
:+1:

PR #281, opened 2015-09-17
Ignore vendor tests in code coverage reports

bcrowe:
I think this makes sense as the default config.


markstory:
Seems like a good idea to me as well.

PR #284, opened 2015-09-21
2 spaces for yml,xml,json,ctp,css in editorconfig file

cake17:


.editorconfig
```diff
@@ -13,6 +13,6 @@ trim_trailing_whitespace = true
 [*.bat]
 end_of_line = crlf
 
-[*.yml]
+[*.{yml,xml,json,ctp,css}]
```
dereuromark: Why?

cake17: I thought we could define these standards for HTML and CSS, based on the [standards proposed by the creator of Bootstrap](http://mdo.github.io/code-guide/#html-syntax). I find it more readable with 2 spaces too.

We did not define standards for HTML and CSS so we have different usages, for example in [cake.css](https://github.com/cakephp/app/blob/master/webroot/css/cake.css) : 2 spaces and in [base.css](https://github.com/cakephp/app/blob/master/webroot/css/base.css) : 4 spaces.

I changed it for xml and json because we already adopt it for yml, so why not for these same sort of types?

ionas:
While I really think that BS's grid is great, thats no standards body and do really prefer 4 spaces or tabs.
When you mix CSS with media queries, LESS/Stylus/SCSS it is really easier to read.
Same goes apparently for XML, HTML, etc. not sure about yml as it deeply nests.

I'd though agree that 2 spaces / 4 spaces is not optimal. Take a look at whatever CodeKit.app outputs out of LESS to see what many devs are working with (IMHO).

.editorconfig
dereuromark: Hehe, there's where the mess and discussion starts once you abandon proper tab indentation :P

cake17:
ok, thanks for your answer, I wasn't aware that the 2 spaces could cause issues.
So I'll change the 2 spaces in `cake.css`.

ADmad:
Closing to avoid any bike shedding. Lets keep the indentations they way they are.

.editorconfig
cake17: sorry :flushed: I just read the tab/spaces indentation discussion... Last time I propose someting related :smile:

PR #289, opened 2015-09-29
Change min. stability to "stable".

ADmad:
With the stable release of cakephp 3.1 and related projects it's better to use stable packages.


pr0nbaer:
Be careful with this one: `cakephp/migrations:1.4.0` requires `robmorgan/phinx:dev-master`. Since `minimum-stability` and `prefer-stable` are root-only configuration settings (see https://getcomposer.org/doc/04-schema.md#minimum-stability) composer tries to install cakephp/migrations:1.3, because there is still this issue open: https://github.com/cakephp/migrations/issues/144

PR #292, opened 2015-10-17
Pull Request for the css and js files for the PR #7518 in cakephp/cakephp

TheFRedFox:
The (new) cake-core js and css files, which are needed for the pull request [#7518](https://github.com/cakephp/cakephp/pull/7518) in cakephp/cakephp

### DIFF
--- a/webroot/css/cake-core.css
+++ b/webroot/css/cake-core.css
@@ -1,0 +1,3 @@
+.cake-core-hidden {
+  display: none;
+}
--- a/webroot/js/cake-core.js
+++ b/webroot/js/cake-core.js
@@ -1,101 +1,105 @@
-var postLinkFunction = function () {
-    var elements = document.getElementsByClassName('cake-core-postLink');
+(function () {
+    var postLinkFunction = function () {
+        var elements = document.getElementsByClassName('cake-core-postLink');
 
-    var postLinkFunction = function (event) {
-        var formName = this.getAttribute("data-cake-core-form");
-        var confirmMessage = this.getAttribute('data-cake-core-confirm');
-        if (confirmMessage !== null && confirmMessage !== 'undefined') {
-            if (confirm(confirmMessage)) {
-                document[formName].submit();
-                return false;
+        var postLinkFunction = function (event) {
+            var formName = this.getAttribute("data-cake-core-form");
+            var confirmMessage = this.getAttribute('data-cake-core-confirm');
+            if (confirmMessage !== null && confirmMessage !== 'undefined') {
+                if (confirm(confirmMessage)) {
+                    document[formName].submit();
+                    return false;
+                }
+                return true;
             }
-            return true;
+            document[formName].submit();
+            event.returnValue = false;
+            event.preventDefault();
+            return false;
+        };
+
+        for (var i = 0; i < elements.length; i++) {
+            var element = elements[i];
+            element.onclick = postLinkFunction;
         }
-        document[formName].submit();
-        event.returnValue = false;
-        event.preventDefault();
-        return false;
     };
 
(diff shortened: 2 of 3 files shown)


webroot/css/cake-core.css
```diff
@@ -0,0 +1,3 @@
+.cake-core-hidden {
```
markstory: This could go into one of the existing CSS files.

webroot/js/cake-core.js
```diff
@@ -0,0 +1,22 @@
+var elements = document.getElementsByClassName('cake-core-postLink');
```
markstory: This should probably be linked to the layout by default.

webroot/css/cake-core.css
TheFRedFox: Well, as many don't use or rather need to use the cake.css or bake.css file, they definitely need the cake-core.css file so that postLink and maybe other core methods still work. Therefore I would serve it as a separated file.
My opinion. :wink:

webroot/js/cake-core.js
TheFRedFox: Sry .. I read that over hundred of times now .. but I really don't get it, what you mean. :laughing: Sry.

ankr: Should this wait for DOM ready?

markstory: @TheFRedFox I meant that this script file should have a `<script ... ></script>` tag in the default layout so that it is included by default.

TheFRedFox: @markstory Ok got it! :laughing: Yeah I added cake-core.js and cake-core.css in the default layout.

@ankr Yeah, it should wait for DOM ready. As it is not that trivial to implement it while supporting every browser without native javascript my suggestion was to add the script loading at the end of the html document. But I found some source code to support every browser:
http://javascript.info/tutorial/onload-ondomcontentloaded
I added this into the js file now. But unfortunately I can't test it for older browsers.

webroot/js/cake-core.js
```diff
@@ -0,0 +1,101 @@
+var postLinkFunction = function () {
```
ankr: This file is introducing global variables. Since it's self contained you could just wrap the whole file in an IIFE `(function () { ... }());`

ionas: I this JS should NOT be in the core and instead there should be a basic XHR call which loads a delete view (GET) which has a confirm button). If JS is disabled the whole framework will work unobtrusively and the only JS required is a general js handler that loads get requests as XHR HTTP requests instead of regular HTTP requests (a very few lines that are useful otherwise for RAD as well).

TheFRedFox: @ankr Yeah, you are right. I will change that.
@ionas Sorry, but I don't really get your point. The functionality with this script and stylesheet is already implemented in CakPHP 3. These are changes to support CSP (Content Security Policy): Separating inline js and css in separate files. So this is the same functionality but in separated loaded files.

chinpei215:
Closing, as this approach is not secure. See [my comment](https://github.com/cakephp/cakephp/pull/7518#issuecomment-209780368)
In addition, this script seems to be copied from [here](http://javascript.info/tutorial/onload-ondomcontentloaded#the-crossbrowser-domcontentloaded-handling-code). The license is not suitable for us.
> The content of this site is available under the terms of CC BY-NC-SA.

TheFRedFox:
> In addition, this script seems to be copied from here. The license is not suitable for us.

Ok, that is indeed embarassing for me. :laughing: Sry, for that. In future I will definitely save the websites to have a closer look after I quickfixed or rather implement something to see if it would work.
Thx for finding that out. And yes, I am quite sure, that this is the website I used. :wink:

PR #295, opened 2015-11-07
Add tests for the PagesController.

markstory:
These tests used to be in cakephp/cakephp but the code is here, and not there.


lorenzo:
Good idea

markstory:
It also acts as an in-repo example of how to use the IntegrationTestCase which is a nice side benefit.

PR #301, opened 2015-11-25
Update phpunit.xml.dist to include code coverage white list

timhoeppner:
When attempting to use the default phpunit configuration and generating code coverage I received the error "No whitelist configured, no code coverage will be generated". This change adds the the TestCase folder to the white list so that code coverage can be successfully generated.

phpunit 5.0.9

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -35,5 +35,8 @@
         <blacklist>
             <directory suffix=".php">./vendor/</directory>
         </blacklist>
+        <whitelist>
+            <directory suffix=".php">./tests/TestCase/</directory>
+        </whitelist>
     </filter>
 </phpunit>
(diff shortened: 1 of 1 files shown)


markstory:
Thanks :+1:

phpunit.xml.dist
```diff
@@ -35,5 +35,9 @@
         <blacklist>
             <directory suffix=".php">./vendor/</directory>
         </blacklist>
+        <whitelist>
+            <directory suffix=".php">./tests/TestCase/</directory>
```
dereuromark: Why whitelisting the Tests?

timhoeppner: Sorry, you're right, makes no sense. The source directory would be sufficient.

PR #303, opened 2015-12-04
Add comment on cache duration override in bootstrap

hakito:
There is no information in the app.default.php that the values specified for the core cache files will be overridden in bootstrap.php

I was very confused, that there was no hint in the cake book, nor in the comments. I had to step debug through the whole caching to find out why the settings in the bootstrap.php are overwritten.


bcrowe:
Thanks!

PR #306, opened 2015-12-07
Correct the layouts directory path in docblock.

joshwyrick:
Changed "/app/Template/Layout" to "/src/Template/Layout".


lorenzo:
Thanks!

PR #308, opened 2015-12-09
Update Installer.php

glensc:
no need to call octdec to extract permission bits with substr

[stat(2)](http://linux.die.net/man/2/stat) last three octets are 777 (aka user-group-other),
the other bits don't matter (i.e 1000 for o+t and 2000 for g+s, etc)


markstory:
Thanks :+1:

PR #309, opened 2015-12-15
minimum version for 3.2 will be 5.5

antograssiot:
and chronos needs 5.5.8


HavokInspiration:
The build failure makes me wonder if we need to do such a modification to the composer file for the migrations plugin.

PR #311, opened 2015-12-21
Add hint for fullBaseUrl in CLI environments

hmic:
Whenever a URL is generated from within a shell context, the fullBaseUrl cannot be known (automatically), like from the $_SERVER['HOST_NAME'],
which leads to problems when e.g. sending emails...
The information and documentation on how to "fix" this is hard to find and hinting in the configuration file for this specific case seems reasonable.

### DIFF
--- a/config/bootstrap_cli.php
+++ b/config/bootstrap_cli.php
@@ -23,7 +23,7 @@ use Cake\Core\Plugin;
 
 // Set the fullBaseUrl to some doubtful/useful default as it cannot be known
 // from e.g. $_SERVER['HOST_NAME'] with no webserver involved (cli-sapi).
-// This is especially very usefull when sending email from the shell.
+// This is especially very usefull when sending email from shells.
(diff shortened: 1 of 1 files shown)


config/bootstrap_cli.php
```diff
@@ -21,6 +21,11 @@
  * be put here.
  */
 
+// Set the fullBaseUrl to some doubtful/useful default as it cannot be known
+// from e.g. $_SERVER['HOST_NAME'] with no webserver involved (cli-sapi).
+// This is especially very usefull when sending email from the shell.
```
dereuromark: typo useful

hmic: useful - single l - is correct! I did not know myself till some minuts ago ;-)

dereuromark: That is exactly why I pointed it out - you got 2 in there.. :D

Issue #310, opened 2015-12-18
Output an empty line when cake.bat is used


junichi11:
I think that an empty line should not be output at the top of output lines if there is no problem.
Because it has to be checked (at least in java) when command list is shown as XML. i.e. The output cannot be used directly...

Thanks.

markstory:
Could you send a pull request to correct the issue? I don't have a windows machine to work with.

junichi11:
Sure :)

PR #312, opened 2015-12-21
Remove an empty line output

junichi11:
PR for #310 

Prevent to output an empty line before the XML declaration when a command is run using `--xml` option.

### DIFF
--- a/bin/cake.bat
+++ b/bin/cake.bat
@@ -17,7 +17,6 @@
 
 :: In order for this script to work as intended, the cake\console\ folder must be in your PATH
 
-@echo.
 @echo off
 
 SET app=%0


markstory:
Thanks :+1:

dereuromark:
Do we have to backport this to 2.x? ( https://github.com/cakephp/cakephp/blob/2.7/lib/Cake/Console/cake.bat )

bin/cake.bat
```diff
@@ -17,7 +17,6 @@
 
 :: In order for this script to work as intended, the cake\console\ folder must be in your PATH
```
dereuromark: I think we should remove this comment in both major versions as it is highly misleading and wrong.

PR #314, opened 2015-12-26
Use svg badges

josegonzalez:
- Switch from service providers to shields.io
- Use svg instead of png
- Use more modern-looking flat-square badges


dereuromark:
:+1:

Issue #43, opened 2014-03-24
Setting debug = false causes error, path not found


spiliot:
I don't think this is a cake core bug

Setup: 
- Follow instructions for manual installation in `/var/www/test`
- Copy `App/Config/app.default.php` to `app.php`

Visiting `localhost/test` works fine.

Setting `debug => false` in `app.php` causes

```
Not Found
Error: The requested address '/test/' was not found on this server.
```

tripatabhishek:
A very simple solution. 
1. Go to routes.php under config 
2. change this line 
     $routes->connect('/', ['controller' => 'Pages', 'action' => 'display', 'home']); 
      to 
     $routes->connect('/', ['controller' => 'yourdesiredcontroller', 'action' => 'yourdesired action']); 
      and that's all.. 
      enjoy........

haulsoft:
In Cakephp 3.0, comment out these lines in src/Template/Pages/home.ctp:
// if (!Configure::read('debug')):
//     throw new NotFoundException();
// endif;

kings360:
I have changed the controller in the route file it works if debug is set as true , but if debug is set as false again it shows error
Error: The requested address '/' was not found on this server.

epixilog:
Well I guess this is so much simple to do such a long conversation : 

On the home.ctp file you have : 

if (!Configure::read('debug')):
    throw new NotFoundException();
endif;

So when you put debug on false (on app.php file) your view throw this exception. So even delete or comment those lines to not have your sweet error page :smile: 
Also your exception isn't supposed to throw too many details as you put debug at false (so you're supposed to be on production)

Hai!

clthck:
@ADmad , @haulsoft , @epixilog  thnx guys. I tried to figure it out a whole day. this is so simple and maybe a stupid issue.

isaacrajaei:
@haulsoft If by getting read of these lines problem solves then why throw new NotFoundException() is there?
In that case you don't need to even input "use Cake\Network\Exception\NotFoundException;" in the view.

@ionas Kindly can you share with us how could you manage this error. (Obviously not by commenting  the scrips)

ADmad:
@isaacrajaei The default homepage shows potentially sensitive environment info that's why it shouldn't be visible in production (that is with debug off), hence the exception. You are supposed to replace it with your own content. The [error message](https://github.com/cakephp/app/blob/master/src/Template/Pages/home.ctp#L24) has already been updated to indicate that.

[...]

PR #319, opened 2016-01-06
Update home.ctp

ionas:
Please review the error message itself, written to errors.log, flagged as warning.
Could you think of a more concise English sentence pointing at that controller/action/template?
Maybe Pages::display/home ?

Ref: https://github.com/cakephp/app/issues/43#issuecomment-38471491
Ref: https://github.com/cakephp/app/issues/43#issuecomment-38473872

Note: I'd actually include an error message, because where is the security issue here with being transparent?

### DIFF
--- a/src/Template/Pages/home.ctp
+++ b/src/Template/Pages/home.ctp
@@ -22,7 +22,7 @@ use Cake\Network\Exception\ForbiddenException;
 $this->layout = false;
 
 if (!Configure::read('debug')):
-    Log::write('warning', 'Pages::display(\'home\'); is only accessible if debug is set to true.');
+    Log::write('warning', 'PagesController::display(\'home\'); is only accessible if debug is set to true.');
     throw new ForbiddenException();
 endif;
 
--- a/src/Template/Pages/home.ctp
+++ b/src/Template/Pages/home.ctp
@@ -22,7 +22,7 @@ use Cake\Network\Exception\ForbiddenException;
 $this->layout = false;
 
 if (!Configure::read('debug')):
-    Log::write('warning', 'Pages/home is only accessible if debug is set to true.');
+    Log::write('warning', 'Pages::display(\'home\'); is only accessible if debug is set to true.');
     throw new ForbiddenException();
 endif;
 
--- a/src/Template/Pages/home.ctp
+++ b/src/Template/Pages/home.ctp
@@ -16,12 +16,14 @@ use Cake\Cache\Cache;
 use Cake\Core\Configure;
 use Cake\Datasource\ConnectionManager;
 use Cake\Error\Debugger;
-use Cake\Network\Exception\NotFoundException;
+use Cake\Log\Log;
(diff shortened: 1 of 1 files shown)


bravo-kernel:
How about adding a (prominent) warning to the `home.ctp` page itself as well to inform new users about this 🕳?

ADmad:
> How about adding a (prominent) warning to the home.ctp page itself as well to inform new users about this?

:+1:  since most newbies wouldn't check the logs either. So better to try and make them aware while the page is accessible :)

lorenzo:
Why not just show the page minus the status checks?

ionas:
@lorenzo with an info box saying "Set debug to true, to see setup status check results"?

ionas:
Ref: https://github.com/cakephp/app/pull/320

bravo-kernel:
Please note that **this page will not be shown** when you turn off debug mode. To disable this security measure remove the NotFoundException in src/Template/Pages/home.ctp.

ionas:
@bravo-kernel well I tend to agree with @lorenzo - would you be so kind and adapt to that? E.g. a notification that status check results will only be shown when debug is turned on? Is that a good compromise?

bravo-kernel:
Unless I am totally misreading the intended end-result... not really and I feel backed by [these arguments](https://github.com/cakephp/app/pull/225): no page at all in non-debug mode by default.

ionas:
I can see reasons for both. I just need a decision and then change it to whatever fits.
@ADmad seems to also have an opinion.
I tend to agree with @AD7six that I'd like this site to not be accessible if the app is set into production mode. By doing so there is no need to remove the page (I usually rename it to cake.ctp or setup.ctp).

jadb:
@ionas, isn't renaming the same as removing? 

IMO, `home.ctp` should just return the usual 404 when not in debug mode. 

Maybe better documenting it in the book if it's to help new users?

ionas:
@jadb what I mean is, I keep the file around for testing the setup in an convenient way and have another default homepage. The file should then still not be accessible unless debug is turned on.

dereuromark:
So how about https://github.com/cakephp/app/pull/320/files with **always** display that error message?
This should make it perfectly clear that deploying this default home.ctp is pointless.

bravo-kernel:
I don't think that PR would help people using cakebox or similar bridged vm @dereuromark since 127.0.0.1 would not be applicable and thus... white page again.

dereuromark:
I just said: ALWAYS display, so no need for the env check :) I will refactor.

markstory:
Closing in preference of #320

ionas:
* Sorry but this PR also wrote a warning. And set HTTP 403 instead of 404. If both are not deemed a good idea, okay.
* What about @lorenzo wish to render the page without the stats?
* What about the general wish to add a note that in case debug is turned off this page will vanish?

markstory:
If the page doesn't leak information I don't see why we couldn't remove the exception.

PR #320, opened 2016-01-06
Make it clearer for beginners to replace home.ctp

dereuromark:
This is making people trip every week.
Especially beginners are not aware and think this is a bug.
The message for dev only should help here.


lorenzo:
As I said in another ticket, I think the most intuitive thing for beginners is having the same page, with no extra info... just the splash screen.

dereuromark:
How about always displaying the message. This makes it super-clear for everyone that you need to replace the file prior to going live with it. The extra message is no security issue IMO.

markstory:
This is an incremental improvement that might help people. Making a 'safe' splash page is a great idea though.

PR #321, opened 2016-01-06
Make it even clearer for beginners to replace home.ctp

bravo-kernel:
Incremental improvement to help out new users even better by informing them through newly added `ctp-warning` splash screen row. Refs https://github.com/cakephp/app/pull/319#issuecomment-169300334

really-help

### DIFF
--- a/webroot/css/cake.css
+++ b/webroot/css/cake.css
@@ -319,6 +319,11 @@ pre {
     margin-top:50px;
 }
 
+.checks.ctp-warning {
+    background-color: #ff5555;
+    padding-bottom: 10px;
+}
+
 .checks.url-rewriting {
     background-color: #F0F0F0;
     display: none;
--- a/src/Template/Pages/home.ctp
+++ b/src/Template/Pages/home.ctp
@@ -47,6 +47,11 @@ $cakeDescription = 'CakePHP: the rapid development php framework';
     </header>
     <div id="content">
         <div class="row">
+            <div class="row">
+                <div class="columns large-12 ctp-warning checks">
+                    <p>Please be aware that this page will not be shown if you turn off debug mode unless you disable the NotFoundException in src/Template/Pages/home.ctp.</p>
+                </div>
+            </div>
             <?php Debugger::checkSecurityKeys(); ?>
             <div id="url-rewriting-warning" class="columns large-12 url-rewriting checks">
                 <p class="problem">URL rewriting is not properly configured on your server.</p>


dereuromark:
I would have thought that the merged exception message suffices.

bravo-kernel:
The exception requires users to look in the logs for the cause right? This informs them beforehand which feels even more helpful to me. Off to :sleeping: now.

dereuromark:
I am still not sure who is actually deploying this boilerplate demo homepage on a server.
I mean, how useless is that? A website's most important (and only) start page is `/`.
Thus this would always be the first one to completely be written from scratch ;)
And even if you do, you will know who why you screwed up badly. But meeeh.

jhli:
I like it.

josegonzalez:
@dereuromark in my cake2 book, the first thing I do is show people how to deploy their apps, even before they make code changes, because I think continuous deployment and testing are so important. If you do that with a CakePHP app, the frontpage is a 404. Which then I have to explain why thats the case and instruct people to change it. So yes, I'm totally in favor of this change.

webroot/css/cake.css
```diff
@@ -319,6 +319,11 @@ pre {
     margin-top:50px;
 }
 
+.checks.ctp-warning {
+    background-color: #ff5555; #bb0000;	#f2dede;
```
markstory: What's the deal with 3 background colors?

bravo-kernel:
Totally looked over the color codes, removed and rebased mr. @markstory,

ionas:
Thank you!

ADmad:
I'll just leave this here: http://stackoverflow.com/questions/36116483/please-be-aware-that-this-page-will-not-be-shown-if-you-turn-off-debug-mode-unle

lorenzo:
:(

lorenzo:
Can remove it?

markstory:
Lets leave it in. At least with the current state people who read the page won't go to stackoverflow.

lorenzo:
Maybe change the color to something that does not look as an error?

bravo-kernel:
Never thought I would reach the point where I would not be able to empathize with the new user anymore :scream:

jhli:
• Yellow/orange instead of red as it is a notice/warning?
• "... Until you replace View/Pages/home.ctp"?

Anyway this won't help the user @Admad's SO post. For those all we can do is explaining dipatching/routing/mvc and posdibly .htaccess and/or how/where/if to place third party public standalone webapps like phpmyadmin; all stuff for the docs.

markstory:
@lorenzo Yellow might work better.

PR #328, opened 2016-01-17
Add support for reading commonly changed config from env vars

### LINKED ISSUE #224: RFC: Adding support for `.env` files
This is something that could help local developers configure their setups much quicker. With the added support for DSNs in the core, it is now quite easy to get setup. 

## Why

This isn't a new way to configure applications. Many of us are already taking advantage of this in production through process managers such as Circus, SupervisorD, and Upstart. As well, all popular webservers support the usage of environment variables (`SetEnv` in Apache, and various ways with Nginx). In production at least, environment variables allow the changing of configuration *without* a deploy, which can be a godsend if, for example, your database goes away and you need to swap to a backup.

The other issue I'd like to tackle is different development environments for teams. It's annoying to have to change hardcoded config when moving from one database config to another (we don't all have `my_app:secret` setup), and having a `.env` file can obviate the need for people stepping over each other while changing config.

## How

We can use [josegonzalez/php-dotenv](https://github.com/josegonzalez/php-dotenv), a library I maintain, or [vlucas/phpdotenv](https://github.com/vlucas/phpdotenv), which is used in Laravel. Both solve similar problems in different ways. The advantage with the former is that we have more control over features we want/need, though to be honest both would work adequately.

Using my library, we can add the following to our `bootstrap.php` file:

```php
try {
    josegonzalez\Dotenv\Loader::load([
        'filepath' => __DIR__ . DS . '.env',
        'toServer' => true,
        'skipExisting' => ['toServer'],
        'raiseExceptions' => true
    ]);
} catch (InvalidArgumentException $e) {
    // do nothing in case the file doesn't exist
}
```

Here is a sample `.env` file that I use for `vagrant` based installations:

```shell
export APP_NAME=app

export DEBUG=2
export SECURITY_SALT="DYhG93b0qyJfIxfs2guVoUubWwvniR2G0FgaC9mi"
export SECURITY_CIPHER_SEED="76859309657453542496749683645"

export DATABASE_URL="mysql://my_app:secret@localhost/my_app?encoding=utf8&timezone=UTC&cacheMetadata=true&quoteIdentifiers=false&persistent=false"
export DATABASE_TEST_URL="mysql://my_app:secret@localhost/my_app?encoding=utf8&timezone=UTC&cacheMetadata=true&quoteIdentifiers=false&persistent=false"

export CACHE_URL="file:///CACHE/?prefix=APP_NAME_&duration=DURATION"
export CACHE_CAKE_CORE_URL="file:///CACHE/persistent/?prefix=myapp_cake_core_&duration=DURATION&serialize=true"
export CACHE_CAKE_MODEL_URL="file:///CACHE/models/?prefix=myapp_cake_model_&duration=DURATION&serialize=true"

export LOG_DEBUG_URL=file:///LOGS/?levels[]=notice&levels[]=info&levels=[]=debug
export LOG_ERROR_URL=file:///LOGS/?levels[]=warning&levels[]=error&levels[]=critical&levels[]=alert&levels[]=emergency

export EMAIL_URL="mail://user:secret@localhost:25/?client=null&timeout=30&tls=null"
```

I do a few replacements in my `app.php` file like so:

```php
'_cake_model_' => [
    'url' => str_replace(
        ['/CACHE/', 'DURATION'],
        [CACHE, '+2 minutes'],
        env('CACHE_CAKE_MODEL_URL')
    ),
],
```

I can likely make the library smarter and have it do it's replacements not only from `bash` but also from PHP constants or existing environment variables.

## Errata

We'd need to add documentation on how `.env` files work. I have quite a bit of it in the readme for my project, but we'd want to add this to the main docs as well.

Note that this would maybe increase the complexity of setting up an app. We don't want to promote a `.env` file in production, but we do want to show users how to configure apps in production. Something to consider.

Developers might think this is something we've stolen from other frameworks, though this sort of thing has been around for quite a while. Rails has had automatic support for DSNs in environment variables since at least 3.x, and the `friendsofcake/app-template` project has used them for at least a year. As well, most/all of the core team has used this sort of functionality in the past, though maybe not with a php library to support their dev cycle.

Does adding this break BC? We've not yet made a release of the `app` composer project. Should we wait until a 3.1 CakePHP release before doing this? Or are we allowed to make changes like this without waiting for a major release.

Finally, does everyone desire this change? I know this makes it easy for me to write my book, and probably makes tutorial writing slightly easier, but maybe this isn't right for our users. Thoughts?

josegonzalez:
Also added App.default_locale.

Closes #224

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -85,7 +85,7 @@ return [
         'default' => [
             'className' => 'File',
             'path' => CACHE,
-            'url' => env('CACHE_DEFAULT_URL', null),
+            'url' => env('CACHE_DEFAULT_URL', ''),
         ],
 
         /**
@@ -99,7 +99,7 @@ return [
             'path' => CACHE . 'persistent/',
             'serialize' => true,
             'duration' => '+2 minutes',
-            'url' => env('CACHE_CAKECORE_URL', null),
+            'url' => env('CACHE_CAKECORE_URL', ''),
         ],
 
         /**
@@ -114,7 +114,7 @@ return [
             'path' => CACHE . 'models/',
             'serialize' => true,
             'duration' => '+2 minutes',
-            'url' => env('CACHE_CAKEMODEL_URL', null),
+            'url' => env('CACHE_CAKEMODEL_URL', ''),
         ],
     ],
 
@@ -182,7 +182,7 @@ return [
             'password' => 'secret',
             'client' => null,
             'tls' => null,
-            'url' => env('EMAILTRANSPORT_DEFAULT_URL', null),
+            'url' => env('EMAILTRANSPORT_DEFAULT_URL', ''),
         ],
     ],
 
@@ -250,7 +250,7 @@ return [
              */
             //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
             
-            'url' => env('DATABASE_URL', null),
+            'url' => env('DATABASE_URL', ''),
         ],
 
         /**
@@ -271,7 +271,7 @@ return [
             'quoteIdentifiers' => false,
             'log' => false,
             //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
-            'url' => env('DATABASE_TEST_URL', null),
+            'url' => env('DATABASE_TEST_URL', ''),
         ],
     ],
 
@@ -284,14 +284,14 @@ return [
             'path' => LOGS,
             'file' => 'debug',
             'levels' => ['notice', 'info', 'debug'],
-            'url' => env('LOG_DEBUG_URL', null),
+            'url' => env('LOG_DEBUG_URL', ''),
         ],
         'error' => [
             'className' => 'Cake\Log\Engine\FileLog',
             'path' => LOGS,
             'file' => 'error',
             'levels' => ['warning', 'error', 'critical', 'alert', 'emergency'],
-            'url' => env('LOG_ERROR_URL', null),
+            'url' => env('LOG_ERROR_URL', ''),
         ],
     ],
 
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -9,7 +9,7 @@ return [
      * Development Mode:
      * true: Errors and warnings shown.
      */
-    'debug' => true,
+    'debug' => env('DEBUG', true),
 
     /**
      * Configure basic information about the application.
@@ -37,7 +37,8 @@ return [
      */
     'App' => [
         'namespace' => 'App',
-        'encoding' => 'UTF-8',
+        'encoding' => env('APP_ENCODING', 'UTF-8'),
+        'default_locale' => env('APP_DEFAULT_LOCALE', 'en_US'),
         'base' => false,
         'dir' => 'src',
         'webroot' => 'webroot',
@@ -62,7 +63,7 @@ return [
      *   You should treat it as extremely sensitive data.
      */
     'Security' => [
-        'salt' => '__SALT__',
+        'salt' => env('SECURITY_SALT', '__SALT__'),
     ],
 
     /**
@@ -84,6 +85,7 @@ return [
         'default' => [
             'className' => 'File',
             'path' => CACHE,
+            'url' => env('CACHE_DEFAULT_URL', null),
         ],
 
         /**
@@ -97,6 +99,7 @@ return [
             'path' => CACHE . 'persistent/',
             'serialize' => true,
             'duration' => '+2 minutes',
+            'url' => env('CACHE_CAKECORE_URL', null),
         ],
 
         /**
@@ -111,6 +114,7 @@ return [
             'path' => CACHE . 'models/',
             'serialize' => true,
             'duration' => '+2 minutes',
+            'url' => env('CACHE_CAKEMODEL_URL', null),
         ],
     ],
 
@@ -178,6 +182,7 @@ return [
             'password' => 'secret',
             'client' => null,
             'tls' => null,
+            'url' => env('EMAILTRANSPORT_DEFAULT_URL', null),
         ],
     ],
 
@@ -244,6 +249,8 @@ return [
              * which is the recommended value in production environments
              */
             //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
+            
+            'url' => env('DATABASE_URL', null),
         ],
 
         /**
@@ -264,6 +271,7 @@ return [
             'quoteIdentifiers' => false,
             'log' => false,
             //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
+            'url' => env('TEST_DATABASE_URL', null),
         ],
     ],
 
@@ -276,12 +284,14 @@ return [
             'path' => LOGS,
             'file' => 'debug',
             'levels' => ['notice', 'info', 'debug'],
+            'url' => env('LOG_DEBUG_URL', null),
         ],
         'error' => [
             'className' => 'Cake\Log\Engine\FileLog',
             'path' => LOGS,
             'file' => 'error',
             'levels' => ['warning', 'error', 'critical', 'alert', 'emergency'],
+            'url' => env('LOG_ERROR_URL', null),
         ],
     ],
 
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -85,7 +85,7 @@ return [
         'default' => [
             'className' => 'File',
             'path' => CACHE,
-            'url' => env('CACHE_DEFAULT_URL', ''),
+            'url' => env('CACHE_DEFAULT_URL', null),
         ],
 
         /**
@@ -99,7 +99,7 @@ return [
             'path' => CACHE . 'persistent/',
             'serialize' => true,
             'duration' => '+2 minutes',
-            'url' => env('CACHE_CAKECORE_URL', ''),
+            'url' => env('CACHE_CAKECORE_URL', null),
         ],
 
         /**
@@ -114,7 +114,7 @@ return [
             'path' => CACHE . 'models/',
             'serialize' => true,
             'duration' => '+2 minutes',
-            'url' => env('CACHE_CAKEMODEL_URL', ''),
+            'url' => env('CACHE_CAKEMODEL_URL', null),
         ],
     ],
 
@@ -182,7 +182,7 @@ return [
             'password' => 'secret',
             'client' => null,
             'tls' => null,
-            'url' => env('EMAIL_TRANSPORT_DEFAULT_URL', ''),
+            'url' => env('EMAIL_TRANSPORT_DEFAULT_URL', null),
         ],
     ],
 
@@ -250,7 +250,7 @@ return [
              */
             //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
             
-            'url' => env('DATABASE_URL', ''),
+            'url' => env('DATABASE_URL', null),
         ],
 
         /**
@@ -271,7 +271,7 @@ return [
             'quoteIdentifiers' => false,
             'log' => false,
             //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
-            'url' => env('DATABASE_TEST_URL', ''),
+            'url' => env('DATABASE_TEST_URL', null),
         ],
     ],
 
@@ -284,14 +284,14 @@ return [
             'path' => LOGS,
             'file' => 'debug',
             'levels' => ['notice', 'info', 'debug'],
-            'url' => env('LOG_DEBUG_URL', ''),
+            'url' => env('LOG_DEBUG_URL', null),
         ],
         'error' => [
             'className' => 'Cake\Log\Engine\FileLog',
             'path' => LOGS,
             'file' => 'error',
             'levels' => ['warning', 'error', 'critical', 'alert', 'emergency'],
-            'url' => env('LOG_ERROR_URL', ''),
+            'url' => env('LOG_ERROR_URL', null),
         ],
     ],
 
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -38,7 +38,7 @@ return [
     'App' => [
         'namespace' => 'App',
         'encoding' => env('APP_ENCODING', 'UTF-8'),
-        'default_locale' => env('APP_DEFAULT_LOCALE', 'en_US'),
+        'defaultLocale' => env('APP_DEFAULT_LOCALE', 'en_US'),
         'base' => false,
         'dir' => 'src',
         'webroot' => 'webroot',
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -271,7 +271,7 @@ return [
             'quoteIdentifiers' => false,
             'log' => false,
             //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
-            'url' => env('TEST_DATABASE_URL', null),
+            'url' => env('DATABASE_TEST_URL', null),
         ],
     ],
 
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -9,7 +9,7 @@ return [
      * Development Mode:
      * true: Errors and warnings shown.
      */
-    'debug' => env('DEBUG', true),
+    'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),
 
     /**
      * Configure basic information about the application.
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -182,7 +182,7 @@ return [
             'password' => 'secret',
             'client' => null,
             'tls' => null,
-            'url' => env('EMAILTRANSPORT_DEFAULT_URL', ''),
+            'url' => env('EMAIL_TRANSPORT_DEFAULT_URL', ''),
         ],
     ],
 
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -97,7 +97,7 @@ mb_internal_encoding(Configure::read('App.encoding'));
  * Set the default locale. This controls how dates, number and currency is
  * formatted and sets the default language to use for translations.
  */
-ini_set('intl.default_locale', Configure::read('App.default_locale'));
+ini_set('intl.default_locale', Configure::read('App.defaultLocale'));
 
 /**
  * Register application error and exception handlers.
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -97,7 +97,7 @@ mb_internal_encoding(Configure::read('App.encoding'));
  * Set the default locale. This controls how dates, number and currency is
  * formatted and sets the default language to use for translations.
  */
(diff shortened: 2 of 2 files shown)


config/app.default.php
```diff
@@ -264,6 +271,7 @@
             'quoteIdentifiers' => false,
             'log' => false,
             //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
+            'url' => env('TEST_DATABASE_URL', null),
```
cake17: Maybe `DATABASE_TEST_URL` would be more consistent with other configs, and to see directly this is about DATABASE ?

cake17:
cool :+1: on this

markstory:
Looks good to me. :+1:

config/app.default.php
```diff
@@ -37,7 +37,8 @@
      */
     'App' => [
         'namespace' => 'App',
-        'encoding' => 'UTF-8',
+        'encoding' => env('APP_ENCODING', 'UTF-8'),
+        'default_locale' => env('APP_DEFAULT_LOCALE', 'en_US'),
```
josegonzalez: Should this be `default_locale` or `defaultLocale`?

dereuromark: Like className we use camelBacked.

jhli:
Looks awesome. As @davidyell pointed out in the RFC, will there be a docs update?

Also will there be an example env file having the word 'example' in its filename? (and could we renane app.default.php to app.example.php - for the first filename some newbies on irc thought it would get loaded by default).

josegonzalez:
I'll try and prepare some docs later tonight which should take care of your questions. By default this doesn't load anything special, and to do that, you'd need a separate library (which I'll also document).

ADmad:
Shouldn't we prefix all the env vars with `CAKE_` to avoid potentially interfering with other scripts/apps on the same environment?

josegonzalez:
Umm no? No other (useful) framework does that. As an example, rails does not use prefixed env vars.

Can you give me an example where this might be a problem?

jippi:
what code actually read and parse out those `url` keys?

ADmad:
The url constants are pretty cake specific and unlikely to be used used anywhere else but var like `DEBUG` is pretty generic.

josegonzalez:
@jippi the parseDsn code I [submitted to the](https://github.com/cakephp/cakephp/pull/4861) core in the betas.

config/app.default.php
```diff
@@ -178,6 +182,7 @@
             'password' => 'secret',
             'client' => null,
             'tls' => null,
+            'url' => env('EMAILTRANSPORT_DEFAULT_URL', ''),
```
AD7six: To be consistent with underscore usage elsewhere, shouldn't this be `EMAIL_TRANSPORT_DEFAULT_URL`?

AD7six:
> Shouldn't we prefix all the env vars with CAKE_ to avoid potentially interfering with other scripts/apps on the same environment?

@ADmad fwiw I haven't seen that ever be a problem, but if there was any ambiguity you would just do this:

    $ source .env; bin/cake

Or

    $ DEBUG=1 bin/cake

Or similar

In a http context it shouldn't be relevant as the site would either not be using env vars, or it'd be full of:

    server {
        listen 80;
        ...
        fastcgi_param DATABASE_URL mysql://root:root@localhost/my_app;
        etc.
    }

ADmad:
@AD7six Gotcha.

config/app.default.php
tmatsuo: :+1:

config/app.default.php
```diff
@@ -244,6 +249,8 @@
              * which is the recommended value in production environments
              */
             //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
+            
+            'url' => env('DATABASE_URL', ''),
```
ionas: `env()` also prases whatever APACHE (and nginx) supply, doesn't it?

If that is the case we should stick to a prefix to make sure that that stuff did not come from the webserver such as:
`CAKE_*` or `CONF_*`. This will help twofold:

- Make transparent where things came from.
- Make sure there are no accidental clashes between the environment the server sets, and the environment you set in your env file

This is just for the convention out of the box. You can still customize the app config to your desires and utilize other env pseudo-constants.

AD7six: > Make transparent where things came from.

I don't see how using a prefix affects that

> Make sure there are no accidental clashes between the environment the server sets, and the environment you set in your env file.

There would need to be at least one hypothetical clash for doing that to make sense IMO. Using a prefix prevents doing things like this:

    $ DATABASE_URL=xxx; bin/cake
    $ DATABASE_URL=xxx; bin/foo.py
    $ DATABASE_URL=xxx; bin/foo.rb

I don't see a need or benefit to using a prefix - it just complicates things for the default case, to account for an edge case scenario (which could instead use different env vars, or remove them etc.).

ionas: `env()` does some more things: http://book.cakephp.org/3.0/en/core-libraries/global-constants-and-functions.html

In case the environment variables come directly via $\_SERVER I don't see a problem (or rather, the problem is shifted to properly configuring the app server environment on the app server level). In case however there is a .env file things can clash. The convention should be helpful towards non-issues. The case you make above is already out of easy and conventional use.
If you don't like the `CAKE_` prefix, `CONF_` or `APP_` would do okay too. 

Another way to do it is to throw a HTTP 500 if a variable is set via `$_SERVER` and then is overwritten by .env. Hidden clashes are no fun (dangerous, error prone).

config/app.default.php
```diff
@@ -9,7 +9,7 @@
      * Development Mode:
      * true: Errors and warnings shown.
      */
-    'debug' => true,
+    'debug' => env('DEBUG', true),
```
ionas: For instance having such a thing without a prefix is no good.
For the databases I can see the reason to not prefix things *in some edge cases*, but I would by convention I'd still prefix those, too.

config/app.default.php
```diff
@@ -62,7 +63,7 @@
      *   You should treat it as extremely sensitive data.
      */
     'Security' => [
-        'salt' => '__SALT__',
+        'salt' => env('SECURITY_SALT', '__SALT__'),
```
ionas: Again a prefix is good. Etc.

config/app.default.php
AD7six: > In case however there is a .env file

This PR doesn't add any support for parsing/using a `.env` file, the only reference to `.env` is my answer to ADmad's question.

> Another way to do it is to throw a HTTP 500 if a variable is set via $_SERVER and then is overwritten by .env. Hidden clashes are no fun (dangerous, error prone).

If you source a file on the cli, it's because you want to use the env vars defined in that file. It would be weird (not to mention obstructive and irritating) to define an env file causing the app to throw a 500.

Can you give a complete (and realistic) example/scenario of what you mean please - I suspect your objections stem from a misunderstanding.

config/app.default.php
tmatsuo: Good point. I prefer prefixing if the variable is too generic. I can think of many scenarios for the conflicts for the generic `DEBUG`.

ionas:
```
CAKE_DEBUG_FLAG
APP_DEFAULT_LOCALE
CAKE_SECURITY_SALT
CAKE_CACHE_DEFAULT_URL
CAKE_CACHE_PERSISTENT_URL
CAKE_CACHE_MODEL_URL
APP_EMAILTRANSPORT_DEFAULT_URL
APP_DATABASE_URL
APP_DATABASE_TEST_URL
APP_LOG_DEBUG_URL
APP_LOG_ERROR_URL
```
or
```
CAKE_DEBUG_FLAG
CAKE_DEFAULT_LOCALE
CAKE_SECURITY_SALT
CAKE_CACHE_DEFAULT_URL
CAKE_CACHE_PERSISTENT_URL
CAKE_CACHE_MODEL_URL
CAKE_EMAILTRANSPORT_DEFAULT_URL
CAKE_DATABASE_URL
CAKE_DATABASE_TEST_URL
CAKE_LOG_DEBUG_URL
CAKE_LOG_ERROR_URL
```

config/app.default.php
ionas: This PR is however is also meant to allow parsing of .env files like @josegonzalez's Plugin does, isn't it?
Ignore my concerns if you deem them to be irrelevant, anything further can be explained on IRC and is a question on my side then.

tmatsuo: Let me chime in. Here is an example scenario.

Running multiple processes in a Docker container including nginx, php-fpm, etc. I'm passing the env vars outside from Docker. `DEBUG` is too generic and it may affect other process other than CakePHP.

config/app.default.php
josegonzalez: Having already used environment variables in real-world scenarios, I'm going to avoid doing this. It's easy enough to just remove the line in case you *do* mess up your environment, and the naming schema is compatible with how other frameworks do this.

config/app.default.php
josegonzalez: That is the only one that I am *maybe* going to change to `CAKE_DEBUG`. Though honestly I can't think of another piece of software you would use in conjunction with cake where that is an issue.

config/app.default.php
```diff
@@ -84,6 +85,7 @@
         'default' => [
             'className' => 'File',
             'path' => CACHE,
+            'url' => env('CACHE_DEFAULT_URL', ''),
```
ionas: The cache URLs should be consistent ans transparent in naming like `CACHE_DEFAULT_URL`, and `CACHE_PERSISTENT_URL` and `CACHE_MODELS_URL`.

config/app.default.php
ionas: :+1:

config/app.default.php
tmatsuo: BTW, I'm worrying that the debug bit will work as expected. The env variables come as strings.

```php
<?php
$debug = 'false';
if ($debug) {
    echo "No---\n";
}
```

ionas: **IF** it is a problem: `'debug' => strtolower(env('DEBUG')) === 'true' ? true : false,`

tmatsuo: Maybe need to do:

```php
'debug' => (boolean) env('DEBUG', true),
```

ionas: in the twisted PHP world at least in my book `var_dump((boolean) 'false'); // true`

e.g.
```php
<?php
var_dump((boolean) false); // false
var_dump((boolean) 'false'); // true
var_dump((boolean) '0'); // false
var_dump((boolean) 'NULL'); // true
var_dump((boolean) ''); // false

Comment (tmatsuo) [inline config/app.default.php]: Sorry `(boolean)` cast doesn't work. The ternary that @ionas showed might work, but it doesn't work if you set "True". Maybe the following?

```php
<?php

$debug = getenv('DEBUG');

var_dump($debug);

$debug = filter_var($debug, FILTER_VALIDATE_BOOLEAN);

var_dump($debug);

if ($debug) {
    echo "No---\n";
}
```

tmatsuo: In this PR case, it is
```php
'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN);
```

AD7six: `DEBUG` as an env var would be an int - i.e. 0 or 1

    'debug' => (boolean) env('DEBUG', true),

is therefore appropriate imo.

config/app.default.php
josegonzalez: I'm not adding a prefix.

config/app.default.php
josegonzalez: That's pretty fair. I named them this way because the naming is `_cache_model_`, which is pretty silly imo but *shrug* cake amirite?

config/app.default.php
josegonzalez: I was trying to just uppercase the key and then replace the nesting with an underscore.

config/app.default.php
tmatsuo: @AD7six `FILTER_VALIDATE_BOOLEAN` is working very nicely. It converts '0', 'false', 'FALSE', 'False', 'off' to `false`, it converts 'true', 'True', 'on', '1', to `true`. It is clearly better than other options.

AD7six: @tmatsuo indeed it does, but I don't consider DEBUG=off or similar a good practice.

> It is clearly better than other options.

Clearly I disagree =).

tmatsuo: @AD7six What's the downside of using the filter and allow 'off' for being false? I only see the benefit of using the filter.

If you need to set '0' or '1' for debug, it's very confusing. Likewise, if it needs to be exactly 'true' and 'false', it's also confusing.

config/app.default.php
tmatsuo: Re: DATABASE_URL

I'm trying similar change on my side, when I run the following:

```
$ env DATABASE_URL="mysql://USER:PASSWORD@IPADDRESS/DBNAME" bin/cake server
```

I'm getting the following warning:
```
Notice Error: Undefined index: className in [(omitted)...vendor/cakephp/cakephp/src/Datasource/ConnectionManager.php, line 116]
```

and can not connect to the DB.

What's the right syntax for the DATABASE_URL?

Also do you have a plan to add tests for it?

josegonzalez: This is tested in the cakephp core actually. What version of cakephp are you running where this doesn't work?

Also, at least on my shell (bash) `env` prints the current env, it doesn't set an environment variable. What shell are you using?

tmatsuo: @josegonzalez 

I'm using CakePHP 3.1.8.

I figured it out what's going here.

If I only set `DATABASE_URL` (without `DATABASE_TEST_URL`), CakePHP failed to create a connection with the empty url for the test connection. Somehow it can not connect to the DB.

If I set both `DATABASE_URL` and `DATABASE_TEST_URL`, it works just fine.

Is it expected?

josegonzalez: Seems wrong. It should toss out the empty dsn. I'll verify later today, but can you post details of your php env?

tmatsuo: re: env
My `env` is for "running a program in a modified environment". See `man env`.

tmatsuo: @josegonzalez 
Nothing special. I just set `DATABASE_URL`, without setting `DATABASE_TEST_URL`, then I got the warning and CakePHP can not connect.

tmatsuo: @josegonzalez In case you're asking my php version, here is the one.

```
$ php -v
PHP 5.5.9-1ubuntu4.14 (cli) (built: Oct 28 2015 01:34:46) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.3, Copyright (c) 1999-2014, by Zend Technologies
    with Xdebug v2.2.3, Copyright (c) 2002-2013, by Derick Rethans
```

Let me know if you need more

tmatsuo: FWIW, according to my quick debug, it seems like CakePHP try to parse the empty string.

When I change `app.php` to something like this:
```php
'url' => env('DATABASE_TEST_URL', null),
```

It works as expected.

josegonzalez: Interesting, a previous version of this pr used null and I changed it as my testing said that wasn't the way forward. I'll test this later tonight and see what I can dig up.

config/app.default.php
AD7six: @tmatsuo I don't like fuzzy values.

IME it's not normal for a boolean env var to allow fuzzy-values, therefore I don't feel it should be allowed here.

> If you need to set '0' or '1' for debug, it's very confusing.

I think that's only confusing if you don't typically use env vars

> Likewise, if it needs to be exactly 'true' and 'false', it's also confusing.

Luckily nobody suggested that so it doesn't need considering =)

Whatever @josegonzalez says though.

config/app.default.php
jadb: Am against prefixing these constants. They become insanely long in real scenarios.

config/app.default.php
josegonzalez: `0` or `1` are pretty common alternative values  for booleans, and PHP treats them more or less fine (in debug's case, I believe it is completely fine.).

tmatsuo: I can live with that, but what's the downside of using the filter? With the filter, I can easily imagine that it will help certain amount of people from being confused in the future.

config/app.default.php
ADmad: [isset()](https://github.com/cakephp/cakephp/blob/master/src/Core/StaticConfigTrait.php#L104) is used not empty().

config/app.default.php
tmatsuo: One more thing to consider. if `DEBUG` is set to true, your application can potentially leak your secret, right?
Considering that, I think you'd better use the filter.

The `DEBUG` configuration is sensitive. It should default to `false` especially in production environment in my opinion. You choose the `true` default value, probably because it's good for first-time users. That's good, but then you need more safeguard for people.

If anyone set the `DEBUG` env var to `false` (or other potential values indicates false except for `0`), the app can still run with DEBUG=true, which is very dangerous. Their secrets may be publicly available if the app has an error. You can still say RTFM to them, but it's not user friendly.

config/app.default.php
josegonzalez: Looks like you're right. Will change them all to `null`...

tmatsuo:
:+1:

markstory:
Nice work folks :clap:

josegonzalez:
Where should I write docs for this?

markstory:
I think including information on which env vars come out of the box, and how to use env vars in general could [fit well here](http://book.cakephp.org/3.0/en/development/configuration.html#configuring-your-application).

mikehancoski:
Did the Doc get updated ?

josegonzalez:
No, I didn't get around to doing so.

mikehancoski:
Any chance there is an example of using this someplace ?

mikehancoski:
So I found a way to use this I am pretty sure it is the "wrong" way per I am adding code to bootstrap.php but if any one was wondering how to get this working 

First create an .env file and place it in the root of the project 
add lines like this 
```
PARAM_NAME=VALUE
PARAM_NAME2=VALUE2
```

Then (this is the odd/ weird thing I do) add this to the bootstrap.php

```
if(is_file(ROOT . DS . '.env')) {
    $handle = fopen(ROOT . DS . '.env', "r");
    while (($line = fgets($handle)) !== false) {
        $arr = explode('=', $line);
        // stupid line endings
        $arr[1] = trim($arr[1], "\r");
        $arr[1] = trim($arr[1], "\n");
        $arr[1] = trim($arr[1], "\r\n");
        putenv("$arr[0]=$arr[1]");
    }
    fclose($handle);
}
```

Now you can call 
```
evn(PARAM_NAME,'Default value');
```

from for example app.php

 I am pretty sure this is not the best way to do this but, it gets the job done.

luke83:
@mikehancoski #224 should be a start (specially "How" paragraph), waiting ufficial docs.

PR #329, opened 2016-01-18
Master localize homepage

cake17:


ionas:
Thanks for doing this!

src/Template/Pages/home.ctp
```diff
@@ -42,71 +42,71 @@ $cakeDescription = 'CakePHP: the rapid development php framework';
...
-            <h1>Get the Ovens Ready</h1>
+            <h1><?= __d('cake', 'Get the Ovens Ready') ?></h1>
         </div>
     </header>
     <div id="content">
         <div class="row">
             <div class="row">
                 <div class="columns large-12 ctp-warning checks">
-                    <p>Please be aware that this page will not be shown if you turn off debug mode unless you disable the NotFoundException in src/Template/Pages/home.ctp.</p>
+                    <p><?= __d('cake', 'Please be aware that this page will not be shown if you turn off debug mode unless you disable the NotFoundException in src/Template/Pages/home.ctp.') ?></p>
```
dereuromark: I don't think dev warnings should be translatable. We removed them in 3.x from a lot of these developer only things (used to be cake_dev domain).

dereuromark:
I consider almost 99% of those translations dev only and as such should not be translated IMO.

src/Template/Pages/home.ctp
ionas: I do agree that errors and warnings should reside in its own domain. There is no problem then as one can omit translation? Putting them in the cake domain is not good.

ionas:
Just because I do agree with you that software should be English internally, only. That's not the stance of everybody and the world (possibly not too much in Japan or France, an assumption being made here by me)

src/Template/Pages/home.ctp
```diff
@@ -21,10 +21,10 @@ use Cake\Network\Exception\NotFoundException;
 $this->layout = false;
 
 if (!Configure::read('debug')):
-    throw new NotFoundException('Please replace Pages/home.ctp with your own version.');
+    throw new NotFoundException(__d('cake', 'Please replace Pages/home.ctp with your own version.'));
 endif;
 
-$cakeDescription = 'CakePHP: the rapid development php framework';
+$cakeDescription = __d('cake', 'CakePHP: the rapid development php framework');
```
dereuromark: You would need to make all names {n} parts as they are not subject of translation, ever.

src/Template/Pages/home.ctp
```diff
@@ -21,10 +21,10 @@ use Cake\Network\Exception\NotFoundException;
 $this->layout = false;
 
 if (!Configure::read('debug')):
-    throw new NotFoundException('Please replace Pages/home.ctp with your own version.');
+    throw new NotFoundException(__d('cake', 'Please replace Pages/home.ctp with your own version.'));
```
dereuromark: You would need to make all paths dynamic {n} parts as they are not subject of translation, ever.

src/Template/Pages/home.ctp
```diff
@@ -42,71 +42,71 @@ $cakeDescription = 'CakePHP: the rapid development php framework';
...
             <div id="url-rewriting-warning" class="columns large-12 url-rewriting checks">
-                <p class="problem">URL rewriting is not properly configured on your server.</p>
+                <p class="problem"><?= __d('cake', 'URL rewriting is not properly configured on your server.') ?></p>
                 <p>
-                    1) <a target="_blank" href="http://book.cakephp.org/3.0/en/installation.html#url-rewriting">Help me configure it</a>
+                    1) <a target="_blank" href="http://book.cakephp.org/3.0/en/installation.html#url-rewriting"><?= __d('cake', 'Help me configure it') ?></a>
                 </p>
                 <p>
-                    2) <a target="_blank" href="http://book.cakephp.org/3.0/en/development/configuration.html#general-configuration">I don't / can't use URL rewriting</a>
+                    2) <a target="_blank" href="http://book.cakephp.org/3.0/en/development/configuration.html#general-configuration"><?= __d('cake', "I don't / can't use URL rewriting") ?></a>
```
dereuromark: You would need to make all urls {n} parts as they are not subject of translation, ever.

src/Template/Pages/home.ctp
```diff
@@ -42,71 +42,71 @@ $cakeDescription = 'CakePHP: the rapid development php framework';
...
                 </p>
             </div>
             <div class="columns large-5 platform checks">
                 <?php if (version_compare(PHP_VERSION, '5.4.16', '>=')): ?>
-                    <p class="success">Your version of PHP is 5.4.16 or higher.</p>
+                    <p class="success"><?= __d('cake', 'Your version of PHP is 5.4.16 or higher.') ?></p>
                 <?php else: ?>
-                    <p class="problem">Your version of PHP is too low. You need PHP 5.4.16 or higher to use CakePHP.</p>
+                    <p class="problem"><?= __d('cake', 'Your version of PHP is too low. You need PHP 5.4.16 or higher to use CakePHP.') ?></p>
```
dereuromark: Same for versions and alike

dereuromark:
90% of all translations are currently invalid changes, and create a huge overhead for nothing.
They are also subject of constant change.

On top of that history showed that no one ever translated those, they were just useless translations and as such got removed in 3.x. Re-adding them now seems contra-productive.

markstory:
While I can appreciate the effort here, has anyone ever translated any of these strings for a production app? Or is there an interest in translating these strings to provide localized home pages out of the box?

ADmad:
None of these templates are going to be used in production. In general maintaining translation is a PITA and I don't think we want that overhead even if they were provided the first time.

cake17:
Ok I can understand, I just thought we could provide translations in the localized plugin (with maybe a different domain name than cake)
This homepage will surely be deleted, so I'm :-1: for this PR :smile:

PR #337, opened 2016-02-10
More graceful error rendering.

dereuromark:
I think the default error templates should switch from default to error layout.
It took me a while to track down the issue because no logs and no debug output of any kind is emitted in the case when the default.ctp cannot render.
But this can happen actually in almost every app, because the normal dispatching cycle is broken and the view most likely does not have all data available it needs to properly render the default layout.

Result: Always 500 display, even for a simple 400.

In my case a helper call in an element needed sth from a components beforeRender, which never got executed because of the RecordNotFoundException going straight to the rendering.
Thus it is not sane/safe to try to render a layout that by design in most cases cannot safely continue to render.

### DIFF
--- a/src/Template/Error/error400.ctp
+++ b/src/Template/Error/error400.ctp
@@ -2,6 +2,8 @@
 use Cake\Core\Configure;
 use Cake\Error\Debugger;
 
+$this->layout = 'error';
+
 if (Configure::read('debug')):
     $this->layout = 'dev_error';
 
--- a/src/Template/Error/error500.ctp
+++ b/src/Template/Error/error500.ctp
@@ -2,6 +2,8 @@
 use Cake\Core\Configure;
 use Cake\Error\Debugger;
 
+$this->layout = 'error';
+
 if (Configure::read('debug')):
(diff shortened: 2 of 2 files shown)


markstory:
Changing the layout doesn't entirely 'solve' the problem, as the error layout could be updated to depend on application context. But that might be less likely than the default app layout failing.

dereuromark:
It is definitely less likely to fail then the fully enriched default one. If someone wants to take the risk, that person could simply remove this line again.

markstory:
@dereuromark True.

jhli:
Maybe the docs should educate about the recommendation of keepung the error layout dead simple to have a fallback layout that actually does render?

Maybe the docs should also mention the change added here in app as it might not be 'BC', e.g. Devs might assume the past way in future apps.

A note in some cake release might help too (separate section for app).

dereuromark:
Yeah, no navigation, session based auth data display or any other fancy stuff that will most likely break due to missing data coming in from controller/component beforeRender() calls.

PR #338, opened 2016-02-14
minimum-stability stable

ionas:
Ref: https://github.com/cakephp/app/commit/a4406a02dc102c10103fc5b65bab0e549565a1ee
Ref: https://gist.github.com/ionas/a900b900fc98909f7372 (source of the issue was minimum-stability dev)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,6 @@
         "post-install-cmd": "App\\Console\\Installer::postInstall",
         "post-autoload-dump": "Cake\\Composer\\Installer\\PluginInstaller::postAutoloadDump"
     },
-    "minimum-stability": "dev",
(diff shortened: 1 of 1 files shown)


dereuromark:
For new apps this makes probably sense :+1:

dereuromark:
PS: does adding prefer-stable=>true have the same effect?

ionas:
It is already in, and no, it did not (for me, see the gist).

jadb:
I merged because it was to be reverted by @lorenzo's comment on the commit that initially changed it. I believe it was while working on cakephp/chronos and while it wasn't stable yet. 

This setting should be the promoted default for new apps in general (unless one specifically wants dev stability).

PR #341, opened 2016-02-25
Removing echo from a method that returns void

Phillaf:
just seems clearer to me


jhli:
What about changing The methods to not implicilty write to the output buffer.

markstory:
@jhli That is a bigger change that can't be done here.

PR #343, opened 2016-02-27
Use div instead of section.

markstory:
section is supposed to be reserved for generic document sectioning, and
is intended for chapters or numbered sections.

Refs #340


ADmad:
:+1:

PR #349, opened 2016-03-09
Check PHP ver. & req. non-def. PHP exts.

ionas:
Composer Reference: https://github.com/cakephp/cakephp/blob/master/composer.json#L21-L23
HHVM Test: https://3v4l.org/qYMA5

Issue:
https://github.com/cakephp/cakephp/issues/8433#issuecomment-194294226
https://github.com/cakephp/cakephp/issues/8433#issuecomment-194295571

Follow-up PR?:
http://book.cakephp.org/3.0/en/deployment.html info on to deploy utilizing `composer install`.

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -33,9 +33,19 @@ require ROOT . DS . 'vendor' . DS . 'autoload.php';
  */
 require CORE_PATH . 'config' . DS . 'bootstrap.php';
 
+// You can remove this if you are confident that your PHP version is sufficient.
+if (version_compare(PHP_VERSION, '5.5.9') < 0) {
+    trigger_error('You PHP version must be equal or higher than 5.5.9.', E_USER_ERROR);
+}
+
 // You can remove this if you are confident you have intl installed.
 if (!extension_loaded('intl')) {
     trigger_error('You must enable the intl extension to use CakePHP.', E_USER_ERROR);
+}
+
+// You can remove this if you are confident you have mbstring installed.
+if (!extension_loaded('mbstring')) {
+    trigger_error('You must enable the mbstring extension to use CakePHP.', E_USER_ERROR);
 }
 
 use Cake\Cache\Cache;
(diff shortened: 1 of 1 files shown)


config/bootstrap.php
```diff
@@ -33,11 +33,21 @@
  */
 require CORE_PATH . 'config' . DS . 'bootstrap.php';
 
+// You can remove this if you are confident that your PHP version is sufficient.
+if (version_compare(PHP_VERSION, '5.5.9') < 0) {
```
dereuromark: would be nice if we could keep the version DRY, to only have a single source of change if we ever bump it.
Reading from composer file if existent would be cool. Maybe there are better options.

ionas: Stealing my thoughts (in terms of multiple places).
- composer.json
- boostrap
- docs

I completely agree.
However things should read this from boostrap.php somehow. Why? Here it should be as fast as possible.
Reading from the composer.json in boostrap sounds... like slow. :(.
Another option would be to somehow have some pseudo-constant on github ... if that's possible?

It's the same for non default extensions.

ionas: @dereuromark https://github.com/cakephp/app/blob/master/composer.json#L8 might not be required btw because it is already checked by cakephp/cakephp/composer.json
Remove?

markstory: Reading a file and decoding JSON on every request is sub-optimal. The duplication is warranted here, and its not like we are going to _lower_ the version requirement in the future.

markstory:
Nice work. Hopefully this saves us having to field similar questions in the future.

config/bootstrap.php
dereuromark: Well, saving into a common $var here would have micro-optimized the duplication in this file for example.

markstory: People concerned about performance can delete these lines for even bigger speed boosts :smile:

dereuromark: I was mainly referring to a single source of truth inside this page to avoid errors. Performance was not the issue :)

ionas:
@markstory, when composer updating today I saw that `symfony/polyfill-mbstring` gets pulled in as a vendor dependency - I wonder if `!extension_loaded('mbstring')) {` is required then @ https://github.com/cakephp/cakephp/blob/master/composer.json#L23 ?

markstory:
The mbstring polyfill is not one of our direct dependencies so I don't know where it is coming from.

HavokInspiration:
@markstory @ionas Could be the migrations plugin but I am unsure. Phinx has a few symfony dependencies.

PR #351, opened 2016-03-09
Categorised the links with important ones first

daveozoalor:
Felt that the important links should be up first especially the ones that lead the user to help forums and references

### DIFF
--- a/src/Template/Pages/home.ctp
+++ b/src/Template/Pages/home.ctp
@@ -163,28 +163,52 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
                     Our primary goal is to provide a structured framework that enables PHP users at all levels to rapidly develop robust web applications, without any loss to flexibility.
                 </p>
                 <ul>
-                    <li><a href="http://cakefoundation.org/">Cake Software Foundation</a>
-                    <ul><li>Promoting development related to CakePHP</li></ul></li>
-                    <li><a href="http://www.cakephp.org">CakePHP</a>
-                    <ul><li>The Rapid Development Framework</li></ul></li>
-                    <li><a href="http://book.cakephp.org/3.0/en/">CakePHP Documentation</a>
-                    <ul><li>Your Rapid Development Cookbook</li></ul></li>
+                       
+                     <h3>Help and Bug Reports</h3>
+                 
+                    <li><a href="irc://irc.freenode.net/cakephp">irc.freenode.net #cakephp</a>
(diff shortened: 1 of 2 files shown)


src/Template/Pages/home.ctp
```diff
@@ -163,28 +163,52 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
...
+                  
+                    <li><a href="http://www.cakephp.org">CakePHP</a>
+                    <ul><li>The Rapid Development Framework</li></ul></li>
+                    
+                    
+                    
+                    
+                    
+                    
+                    <br/>
```
dereuromark: This looks like invalid HTML
There should be no other HTML elements between `<li>` tags

But apart from this and too much whitespace here, this is a good idea! :+1:

src/Template/Pages/home.ctp
```diff
@@ -163,28 +163,52 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
...
                 </p>
                 <ul>
-                    <li><a href="http://cakefoundation.org/">Cake Software Foundation</a>
-                    <ul><li>Promoting development related to CakePHP</li></ul></li>
-                    <li><a href="http://www.cakephp.org">CakePHP</a>
-                    <ul><li>The Rapid Development Framework</li></ul></li>
-                    <li><a href="http://book.cakephp.org/3.0/en/">CakePHP Documentation</a>
-                    <ul><li>Your Rapid Development Cookbook</li></ul></li>
+                       
+                     <h3>Help and Bug Reports</h3>
```
ionas: This should really be split up into:

- Questions, Getting Help and the Community (more to the top) - also containing the book link possibly.
- Contribution and Bug Reports (more to the bottom)

... maybe helps a bit to get questions into the right channels.

markstory: This isn't valid HTML. h3 cannot be a direct child of `ul`.

daveozoalor: What would be the correct way to do it? What tag should i have used? Do you
suggest i close each <ul> before each <h3> ?

dereuromark: Yes, that is one way to make valid HTML.

markstory: Or put the h3 elements inside an li. Closing the ul will probably be nicer to read though.

markstory:
I did the HTML fixups and some other changes to create #352

Issue #353, opened 2016-03-15
`composer install` Default Permissions


ionas:
When deploying and updating our apps via `composer install` we have run into a few problems regarding permissions.
https://github.com/cakephp/app/blob/master/src/Console/Installer.php#L159

That sets permissions to xx7.
By default however we suggest the permissions to be set to 770 (more secure) or 777 (almost the same like as of now, however if some admin comes along and changes the permissions of the other flag, the cache files will be still readable while xx7 might result into 007).

ionas:
An alternative would be a warning on the console to change this `$worldWritable = 0770;`?
I could setup a PR for that if you think that's sane.

ionas:
Also, what's the benefit of bindec()? instead of a regular *nix bitmask notation?

ADmad:
Not having the files world writable will cause lot of problems to newbies when switching between CLI and webserver. Especially those using packages like WAMP etc.

markstory:
Just change/delete the installer script in your application. The defaults are intended to help new users get going with minimal fuss.

ionas:
@markstory a PR is open. Feel free to close that one if you deem it not any good.

PR #354, opened 2016-03-15
Warn the user about relaxed tmp/logs permissions

ionas:
Also removed use the of `bindec()`.
Can you think of a better worded notice message?

### DIFF
--- a/src/Console/Installer.php
+++ b/src/Console/Installer.php
@@ -156,7 +156,16 @@ class Installer
             }
         };
 
-        $worldWritable = bindec('0000000111');
+        $worldWritable = 007;
+        if ($worldWritable === 007) {
+            $io->writeError(
+                'WARNING: "./tmp" and "./logs" permissions are set to world read- and writable. '
+                . 'You are encouraged to change these xx7 to 700 or 770 before production deployment '
+                . 'in ' . __FILE__ . ' on line #' . (__LINE__ - 5) . ' and then re-run '
+                . '`composer install`.', true, $io::NORMAL
+            );
+        }
+
         $walker($dir . '/tmp', $worldWritable, $io);
         $changePerms($dir . '/tmp', $worldWritable, $io);
         $changePerms($dir . '/logs', $worldWritable, $io);
--- a/src/Console/Installer.php
+++ b/src/Console/Installer.php
@@ -158,11 +158,12 @@ class Installer
 
         $worldWritable = 007;
         if ($worldWritable === 007) {
-            $io->writeError(
-                'WARNING: "./tmp" and "./logs" permissions are set to world read- and writable. '
-                . 'You are encouraged to change these xx7 to 700 or 770 before production deployment '
-                . 'in ' . __FILE__ . ' on line #' . (__LINE__ - 5) . ' and then re-run '
-                . '`composer install`.', true, $io::NORMAL
+            $io->write(
+                'NOTICE: "./tmp" and "./logs" permissions are set to world read- and writable. '
+                . 'Depending on your requirements, you may want to change this before in '
+                . __FILE__ . ' on line #' . (__LINE__ - 5) . ' before production deployment '
+                . 'and then re-run `composer install`.',
+                true, $io::NORMAL
             );
         }
 


ionas:
Ref: https://github.com/cakephp/app/issues/353#issuecomment-196833120

ionas:
Ref: https://github.com/cakephp/app/issues/353#issuecomment-196833120

src/Console/Installer.php
```diff
@@ -156,7 +156,16 @@ public static function setFolderPermissions($dir, $io)
             }
         };
 
-        $worldWritable = bindec('0000000111');
+        $worldWritable = 007;
+        if ($worldWritable === 007) {
+            $io->writeError(
```
jippi: `writeError` won't touch composer exit status, right?

chinpei215:
I am guessing that you have a misunderstand about permissions for webapps.
You might be convinced that 700 is more secure than 777.
However, contrary to your expectations, I would have to say that 700 is more dangerous than 777.

Setting permissions of tmp/logs to 700 implies that the user running the web server has a high level of authority.
If the webapp contains vulnerabilities, such as OS command injection or File upload vulnerability, malicious users can hijack the webapp entirely. 
Because the user can execute dangerous shell commands (e.g. composer), or overwrite existing php files via the webapp.

On the other hand, setting permissions of tmp/logs to 777 implies that user running the web server has a low level of authority.
Even if the webapp contains vulnerabilities, `nobody` user cannot neither execute dangerous shell commands nor overwrite php files.

I think that we shouldn't recommend any permissions.

src/Console/Installer.php
ionas: It does not abort.

ionas:
* Removed the (wrong) security recommendation.
* changed `writeError()` to `write()`.

src/Console/Installer.php
```diff
@@ -156,7 +156,17 @@ public static function setFolderPermissions($dir, $io)
             }
         };
 
-        $worldWritable = bindec('0000000111');
+        $worldWritable = 007;
+        if ($worldWritable === 007) {
```
AD7six: Since the assignment for this is one line above it - this condition will always be true.

src/Console/Installer.php
```diff
@@ -156,7 +156,17 @@ public static function setFolderPermissions($dir, $io)
...
+            $io->write(
+                'NOTICE: "./tmp" and "./logs" permissions are set to world read- and writable. '
+                . 'Depending on your requirements, you may want to change this before in '
+                . __FILE__ . ' on line #' . (__LINE__ - 5) . ' before production deployment '
```
AD7six: Are you really suggesting that users edit this file?

How does a developer creating a new project have anything to do with deployment? :confused:

src/Console/Installer.php
ionas: That's the point. You edit the above permissions and the message won't render.

AD7six: >  You edit the above permissions and the message won't render.

If users are forced to edit the file - they can also just delete the lines generating the notice.

Seems sub-optimal to me (and unnecessary), I don't understand what this has to do with deployment.

ionas: Okay. **Please simply close this PR if you deem it to be unnecessary,** and don't see it - or a variant of it - as an improvement over the status quo.

I went into figuring out why composer install set the other flag to read-write-execute running `composer install` on deployed applications to update their dependencies. I am on IRC for further questions.

TBH the permission flags should only be set when the project is created by composer the first time not every time someone does an update.

ADmad: > TBH the permission flags should only be set when the project is created by composer the first time not every time someone does an update.

This installer is only run when using `composer install` not `composer update`.

ionas: Maybe I missunderstood something but `composer update` will update `composer.lock` whereas `composer install` will simply read `composer.lock` which is exactly what you want after deploying an application source update (frozen vendor dependencies might have changed, or not... fire the command and be save as composer.lock went into the git repository).

ionas:
Closing as there is no support for this.
**Be warned that whatever your security policy and setup is, running `composer install` to update a deployed installation after `git pull` will result into tmp and log files being world read-writable.**

ADmad:
@ionas You are crying wolf, read @chinpei215 comment.

Issue #355, opened 2016-03-16
app.php missing after composer create-project --prefer-dist cakephp/app bookmarker


ionas:
http://book.cakephp.org/3.0/en/tutorials-and-examples/bookmarks/intro.html#getting-cakephp:
> The advantage to using Composer is that it will automatically complete some important set up tasks, such as setting the correct file permissions and creating your config/app.php file for you.

```sh
[09:48 PM]-[vagrant@addressservices]-[/var/www/cakephp] 
$ composer create-project --prefer-dist cakephp/app bookmarker
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
Installing cakephp/app (3.2.2)
  - Installing cakephp/app (3.2.2)
    Downloading: 100%         

Created project in bookmarker
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Installing aura/installer-default (1.0.0)
    Downloading: 100%         

  - Installing cakephp/plugin-installer (0.0.12)
    Downloading: 100%         

  - Installing psr/log (1.0.0)
    Downloading: 100%         

  - Installing mobiledetect/mobiledetectlib (2.8.19)
    Downloading: 100%         

  - Installing aura/intl (1.1.1)
    Downloading: 100%         

  - Installing cakephp/chronos (0.4.8)
    Downloading: 100%         

  - Installing cakephp/cakephp (3.2.5)
    Downloading: 100%         

  - Installing symfony/yaml (v3.0.3)
    Downloading: 100%         

  - Installing symfony/filesystem (v3.0.3)
    Downloading: 100%         

  - Installing symfony/config (v3.0.3)
    Downloading: 100%         

  - Installing symfony/polyfill-mbstring (v1.1.1)
    Downloading: 100%         

  - Installing symfony/console (v3.0.3)
    Downloading: 100%         

  - Installing robmorgan/phinx (v0.5.1)
    Downloading: 100%         

  - Installing cakephp/migrations (1.5.7)
    Downloading: 100%         

  - Installing jakub-onderka/php-console-color (0.1)
    Downloading: 100%         

  - Installing jakub-onderka/php-console-highlighter (v0.3.2)
    Downloading: 100%         

  - Installing dnoegel/php-xdg-base-dir (0.1)
    Downloading: 100%         

  - Installing nikic/php-parser (v2.0.1)
    Downloading: 100%         

  - Installing symfony/var-dumper (v3.0.3)
    Downloading: 100%         

[...]

PR #356, opened 2016-03-17
Ensure post install script runs when creating project.

ADmad:
Refs #355

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
     },
     "scripts": {
         "post-install-cmd": "App\\Console\\Installer::postInstall",
+        "post-create-project-cmd": "App\\Console\\Installer::postInstall",
         "post-autoload-dump": "Cake\\Composer\\Installer\\PluginInstaller::postAutoloadDump"
     },
     "minimum-stability": "dev",


ionas:
Thanks for taking care!

lorenzo:
good stuff

ADmad:
A new release show be done.

lorenzo:
@ADmad go for it!

chinpei215:
@ADmad Wouldn't installer run twice if users are using an old version of composer?

ADmad:
@chinpei215 I checked with composer `1.0.0-beta1` and it does run the script twice. Other than being unexpected having it run twice won't cause any actual problem.

We can remove `"post-install-cmd"` if we want to run the script only when creating the project and not when running composer install or update.

chinpei215:
@ADmad It sounds reasonable to me. If we can remove `"post-install-cmd"`, as positive side-effect, it would also solve #354. But as I am not familiar with composer, my opinion wouldn't be reliable. I misunderstood `composer.lock` as a file to prevent duplicate execution until yesterday :stuck_out_tongue:

ADmad:
@chinpei215 Until yesterday I thought create project was akin to install, but apparently composer says "Meh, you are such a n00b, I'll do an update instead." :stuck_out_tongue: 

@ionas Want to do the honors of removing the the post install hook?

ionas:
I'd rather make sure (newbie) users got a recent composer version. See above.

ionas:
Here is my suggestion: https://github.com/cakephp/app/pull/358
It does assume that composer is recent, thus the above PRs to the docs.

PR #358, opened 2016-03-18
improve composer/installer scripts

ionas:
Could not test this yet as I could not figure out how to composer create-project directly from a git repository non-master branch (instead of through packagist):

 https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,8 +32,9 @@
         }
     },
     "scripts": {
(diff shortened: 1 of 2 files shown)


ionas:
I will close this. If you think it has value you can reopen it.

PR #363, opened 2016-04-05
Make default output time zone easy to configure (once PR #8753 hits)

ionas:
Ref: https://github.com/cakephp/cakephp/pull/8753

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -105,8 +105,8 @@ if (!Configure::read('debug')) {
 date_default_timezone_set('UTC');
 
 /**
- * Sets the default output timezones. 
-**/
+ * Sets the default output timezones.
+ */
 Time::setDefaultOutputTimezone(Configure::read('App.defaultOutputTimezone'));
 FrozenTime::setDefaultOutputTimezone(Configure::read('App.defaultOutputTimezone'));
 
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -57,10 +57,7 @@ use Cake\Core\Plugin;
 use Cake\Database\Type;
 use Cake\Datasource\ConnectionManager;
 use Cake\Error\ErrorHandler;
-use Cake\I18n\Date;
-use Cake\I18n\FrozenTime;
-use Cake\I18n\FrozenDate;
-use Cake\I18n\Time;
+use Cake\I18n;
 use Cake\Log\Log;
 use Cake\Mailer\Email;
 use Cake\Network\Request;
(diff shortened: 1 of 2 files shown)


config/bootstrap.php
```diff
@@ -99,6 +103,15 @@
 date_default_timezone_set('UTC');
 
 /**
+ * Sets the default output timezones.
+ * This output timezone is also being used by TimeHelper.
+**/
```
markstory: Missing a space here. (nitpick).

config/bootstrap.php
```diff
@@ -99,6 +103,15 @@
...
+Time::setDefaultOutputTimezone(Configure::read('App.defaultOutputTimezone'));
+Date::setDefaultOutputTimezone(Configure::read('App.defaultOutputTimezone'));
+FrozenTime::setDefaultOutputTimezone(Configure::read('App.defaultOutputTimezone'));
+FrozenDate::setDefaultOutputTimezone(Configure::read('App.defaultOutputTimezone'));
```
markstory: Is this the right time to do this? I would expect that this would be based on user preferences, and we aren't far enough into the application to access user specific preferences just yet.

config/bootstrap.php
ADmad: Also it should be `*/` not `**/`

config/bootstrap.php
ionas: IMHO its good if you have an application where there are nu user specific preferences but the timezone, locale, date/money/number formats are set globally. It is easy and convenient and you can always later change it or remove the line in bootstrap. IMHO.

config/bootstrap.php
```diff
@@ -99,6 +105,12 @@
 date_default_timezone_set('UTC');
 
 /**
+ * Sets the default output timezones. 
+**/
```
ADmad: Should be `*/`

ADmad:
There's a CS error https://travis-ci.org/cakephp/app/builds/130056442#L253

jhli:
I ll take a look at thus ASAP when doing the docs.

ionas:
This should look good now.

lorenzo:
I was trying to think of an alternative way of doing this. Loading the full of Chronos on each request seems unnecessary.

jhli:
Well an alternative would be to configure read and set in constructor?

lorenzo:
still need to think what the alternative could be

mirko-pagliai:
Merged with my projects, I get this (apache log):
`[Thu Jun 09 10:59:44.694783 2016] [:error] [pid 14134] [client ::1:46010] PHP Fatal error:  Call to undefined method Cake\\I18n\\Time::setDefaultOutputTimezone() in .../config/bootstrap.php on line 110`
CakePHP is updated to 3.2.10.

ADmad:
The changes done in this patch depend on changes done to core's master branch which are not part of 3.2.10.

Revert the changes you have done.

mirko-pagliai:
@ADmad ah, ok. I had imagined something like this.
But it is normal, therefore, that the master branch of the app skeleton contains code not yet included in the latest stable release of CakePHP?

ADmad:
You shouldn't be using master branch of either repo, stick to releases.

PR #364, opened 2016-04-07
Don't use locale parsing by default.

markstory:
While this is a pretty neat feature. It causes new users a bunch of grief as they try to accept SQL date formats in en-US locales.

Refs cakephp/cakephp#8593

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -206,17 +206,15 @@ DispatcherFactory::add('Routing');
 DispatcherFactory::add('ControllerFactory');
 
 /**
- * Enable default locale format parsing.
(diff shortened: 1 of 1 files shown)


josegonzalez:
:)

dereuromark:
We could have kept it commented out as well :)

jhli:
That'd be good and put a link to the book in that docblock?

dereuromark:
Or just a link, it looks like the docs are pretty straight forward on it: http://book.cakephp.org/3.0/en/core-libraries/internationalization-and-localization.html#parsing-localized-datetime-data

jhli:
I ll do that.

PR #366, opened 2016-04-11
make AjaxView extends View instead of AppView

ayman-alkom:
I found this more logical then we are more free in playing with AppView

### DIFF
--- a/src/View/AjaxView.php
+++ b/src/View/AjaxView.php
@@ -17,13 +17,14 @@ namespace App\View;
 use Cake\Event\EventManager;
 use Cake\Network\Request;
 use Cake\Network\Response;
+use Cake\View\View;
 
 /**
  * A view class that is used for AJAX responses.
  * Currently only switches the default layout and sets the response type -
  * which just maps to text/html by default.
  */
-class AjaxView extends AppView
+class AjaxView extends View
 {
 
     /**
--- a/src/View/AjaxView.php
+++ b/src/View/AjaxView.php
@@ -14,9 +14,6 @@
  */
 namespace App\View;
 
-use Cake\Event\EventManager;
-use Cake\Network\Request;
-use Cake\Network\Response;
 use Cake\View\View;
 
 /**


markstory:
Why?

ayman-alkom:
In the case we use the custom view class of BootstrapUI plugin `UIView`, it made conflicts to the debug_kit plugin.
I found that making the AjaxView separated of AppView will be more clear

markstory:
My worry is that we've documented `AppView` as the base class of application views which includes this view class. Not extending `AppView` will mean that custom helpers needed in all views will not be present in `AjaxView`.

ayman-alkom:
> Not extending AppView will mean that custom helpers needed in all views will not be present in AjaxView

I 'm not sure, but I think that we don't use any of the helpers in `AjaxView`.

ADmad:
> I 'm not sure, but I think that we don't use any of the helpers in AjaxView.

`AjaxView` itself doesn't load any helpers but since it extends `AppView` any helpers you load in `AppView` (which is the recommended way) they would be available to `AjaxView` too.

I too don't like like this change. Most people would expect settings done in `AppView` to be able in `AjaxView` too since it only difference is `AjaxView` uses an empty layout.

@ayman-alkom You can make required change in your own app but the official app skeleton should remain the same.

ayman-alkom:
@ADmad I don't know why some one will need FormHelper for example in AjaxView, AjaxView and AppView are completely different  IMO.

I use AjaxView Just for REST API, and don't know if some one using it for the other target 

> You can make required change in your own app but the official app skeleton should remain the same.

Anyone know that :)

ADmad:
> I use AjaxView Just for REST API, and don't know if some one using it for the other target

Most people use AjaxView to return HTML fragments.  What response format are you using for your REST API?  Most ppl would use json or xml and hence JsonView or XmlView not AjaxView.

dereuromark:
Or my Ajax Plugin which has an json powered Ajax View class.

ayman-alkom:
> Most ppl would use json or xml and hence JsonView or XmlView not AjaxView.

yes @ADmad you are right, I forgot JsonView and XmlView, in fact I using them, but  anyway  this change solved my Debugkit issue and other issues related to ajax requests. and you are free consider AjaxView as JsonView must be separated by default or not.

ayman-alkom:
@dereuromark I looked at your [AjaxView](https://github.com/dereuromark/cakephp-ajax/blob/master/src/View/AjaxView.php#L22) and sow it extend the View and not AppView

ADmad:
>  anyway this change solved my Debugkit issue

What debugkit issue?

ayman-alkom:
## In the case it extends View
untitled

## In the case it extends AppView:
my assets comes again every time I view a DebugKit panel, note that I 'm using UIView and this may be able to solved in other way but didn't search more because I found this is the best solution to me.

untitled2

ionas:
I think we should rather offer an `HtmlView` or something long those lines and allow specifying helpers for specific Views, so that you can opt-in/opt-out of setting helpers for `AjaxView`.

You can also create your own XhrAppController and extent your controllers from that AppController which possibly would not load any helpers... (also create an HtmlAppController which does, both derive from AppController, your action holding controllers then from XhrApp-/HtmlAppController) etc.

ADmad:
I'll let @markstory look into the debugkit issue :smile:

markstory:
One way around the DebugKit issue is for DebugKit to provide its *own* AjaxView that userland code can't muck about with.

lorenzo:
Since AjaxView is a class you can own, you can also make it extend whatever you want in your project. I think AppView is the best default.

markstory:
Closing as I feel the current setup is a good default. I will fix the debugkit issue so others don't bump into this specific issue.

PR #367, opened 2016-04-11
Add awesome cakephp list to app template home page

dereuromark:
Needs to be more visible IMO.
Please still don't know about it and fail to not reinvent the wheel or might even chose/follow outdated approaches because of it.

### DIFF
--- a/src/Template/Pages/home.ctp
+++ b/src/Template/Pages/home.ctp
@@ -204,6 +204,10 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
                         <ul><li>For the Development of CakePHP Git repository, Downloads</li></ul>
                     </li>
                     <li>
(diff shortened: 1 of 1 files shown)


bravo-kernel:
+1 on adding the awesome list but this might also be the moment to change to order of this list to something more logical. It feels weird to me that the api comes first, cake site last. I would suggest something like :

1. http://www.cakephp.org
2. https://github.com/cakephp
3. http://book.cakephp.org/3.0/en/
4. http://api.cakephp.org/3.0/
5. https://github.com/FriendsOfCake/awesome-cakephp
6. http://plugins.cakephp.org
7. http://bakery.cakephp.org

dereuromark:
Changing the order should be a separate issue and should not be related to this one IMO. Pickypacking it upon this here only prevents pushing incremental improvements.

PR #369, opened 2016-04-16
Check if debugkit is loaded and suggest possible resolution if it's not.

ADmad:
The error message in logs is not enough. Newbies don't know that they should check logs.


jhli:
:+1:

PR #374, opened 2016-04-27
Render PHP version transparently

ionas:
This will make it easier for newbies to know their eco system in a snap.

### DIFF
--- a/src/Template/Pages/home.ctp
+++ b/src/Template/Pages/home.ctp
@@ -65,9 +65,9 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
             <div class="columns large-12 checks">
                 <h4>Environment</h4>
                 <?php if (version_compare(PHP_VERSION, '5.5.9', '>=')): ?>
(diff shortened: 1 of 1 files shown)


markstory:
Good idea!

src/Template/Pages/home.ctp
```diff
@@ -65,9 +65,9 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
             <div class="columns large-12 checks">
                 <h4>Environment</h4>
                 <?php if (version_compare(PHP_VERSION, '5.5.9', '>=')): ?>
-                    <p class="success">Your version of PHP is 5.5.9 or higher.</p>
+                    <p class="success">Your version of PHP is 5.5.9 or higher (detected <?= phpversion() ?>).</p>
```
ravage84: Nitpicking:
There's a mix of ``PHP_VERSION`` and ``phpversion()``.
Also the  value gets used at least twice. So it could be saved to a variable and reused.

```` php
$phpVersion = phpversion();
````

PR #377, opened 2016-04-27
Cake down to my app

chrisShick:


webroot/.htaccess
```diff
@@ -2,4 +2,6 @@
     RewriteEngine On
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteRule ^ index.php [L]
+    RewriteCond %{HTTP:Authorization} ^(.*)
```
jhli: I ll just say what is good, because the whole PR makes no sense to me.

This RewriteCond is good!

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "cakephp/app",
+    "name": "chrisshick/app",
```
jhli: No.

PR #378, opened 2016-04-30
Add note to ommit periods in database name.

userlond:
https://github.com/cakephp/cakephp/issues/6471


ionas:
Could you add semicolons to the list as FormHelper and composite primary/foreignKeys uses those to create select dropdowns?

PR #379, opened 2016-05-03
[3.2] Including CakePHP official forum link in default home page

andtxr:


ravage84:
:+1:

src/Template/Pages/home.ctp
```diff
@@ -170,6 +170,8 @@ $cakeDescription = 'CakePHP: the rapid development php framework';
                     <ul><li>Everything CakePHP</li></ul></li>
                     <li><a href="http://plugins.cakephp.org">CakePHP plugins repo</a>
```
dereuromark: As a side note: IMO this should be switched to the awesome list
This plugin site is not really maintained (to say the least) and does not really represent CakePHP in a very good way right now.

src/Template/Pages/home.ctp
```diff
@@ -170,6 +170,8 @@ $cakeDescription = 'CakePHP: the rapid development php framework';
...
                     <ul><li>A comprehensive list of all CakePHP plugins created by the community</li></ul></li>
+                    <li><a href="http://discourse.cakephp.org/">CakePHP Official Forum</a>
+                    <ul><li>CakePHP discussion forum</li></ul></li>
                     <li><a href="https://groups.google.com/group/cake-php">CakePHP Google Group</a>
```
dereuromark: You can remove this google group one.

dereuromark:
Why for 3.2 and not for master? 3.2 is the outdated branch.

andtxr:
I do agree about the replacement of plugins site. It looks really outdated.

src/Template/Pages/home.ctp
ravage84: I disagree. it is still in use.

src/Template/Pages/home.ctp
ravage84: Let it in, but add the mentioned link to the awesome list.

src/Template/Pages/home.ctp
dereuromark: No, we deprecated it afaik.

ionas: If its still useful it should come after the awesome list. If you consider it legacy we could mark it as such (Legacy).
Same goes for the CakePHP Google Group in case you want one single forum on discourse.cakephp.org.

lorenzo:
This change is good enough as it is. Comments can be addressed as other PRs

PR #384, opened 2016-05-07
Change varchar to char

### LINKED ISSUE #383: Session table to use char(40) instead of varchar(40)
I think the `sessions.sql` file has been around for several years now, but today I noticed that it was using `varchar(40)` as the primary key.

MySQL (not sure about other DBs) will allow you to use `varchar` as a primary key column, but there are a number of references that say performance can be poor. This of course is relative to the number of rows in the table.

Still, I think this should be changed to just `char(40)` since this column holds the hash key for the session ID and it's length doesn't vary.

Can anyone think of a reason why this would break anything by changing it?

thinkingmedia:
Updates the schema for sessions to use `char(40)` for the primary ID. Also removed the default since this column can not be empty. It has to have the session ID from PHP.

closes #383

### DIFF
--- a/config/schema/sessions.sql
+++ b/config/schema/sessions.sql
@@ -6,7 +6,7 @@
 # MIT License (http://www.opensource.org/licenses/mit-license.php)
 
 CREATE TABLE sessions (
-  id varchar(40) NOT NULL default '',
+  id char(40) NOT NULL,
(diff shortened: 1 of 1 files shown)


markstory:
Thanks :+1:

PR #387, opened 2016-05-26
Micro optimization

lorenzo:
Some time ago, when profiling my app, I noticed that using short class name for the default route class would call the autoloader as many times as you  have routes. In a big app, that have be a couple hundred extra function calls.

This removes that case

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -20,6 +20,7 @@
 
 use Cake\Core\Plugin;
 use Cake\Routing\RouteBuilder;
+use Cake\Routing\Route\DashedRouted;
 use Cake\Routing\Router;
 
 /**
@@ -40,7 +41,7 @@ use Cake\Routing\Router;
  * `:action` markers.
  *
  */
-Router::defaultRouteClass('DashedRoute');
+Router::defaultRouteClass(DashedRoute::class);
 
 Router::scope('/', function (RouteBuilder $routes) {
     /**


markstory:
Neato!

bravo-kernel:
@lorenzo would you be able to share how you profile your application?

config/routes.php
```diff
@@ -20,6 +20,7 @@
 
 use Cake\Core\Plugin;
 use Cake\Routing\RouteBuilder;
+use Cake\Routing\Route\DashedRouted;
```
ADmad: Typo in classname. Extra "d" at the end.

dereuromark: Fixed in https://github.com/cakephp/app/commit/9053056e64f6282641fe36c69d5a20ae8f53ed50

ADmad: Also phpcs complains that this class should be listed last as per alphabetical order :P

dereuromark: Hehe, I let you do the honors of ordering then.

ADmad: Done in 891582d

markstory: Thanks guys :smile:

lorenzo:
@bravo-kernel I can share that after cakefest... just ping me again on tuesday :)

bravo-kernel:
Ping ;)

lorenzo:
I used blackfire.io

They have a CLI application that is combined with a php extension. You do a request in your browser and copy it as curl. Then you replace "curl" with the blackfire CLI command and you get a really nice profiling result of your app.

bravo-kernel:
Thanks for sharing (so quickly) @lorenzo, much obliged

PR #388, opened 2016-05-27
Add PSR7 Skeleton code

markstory:
This updates the app skeleton to use the new `Cake\Http` library instead of using DispatcherFilters. While the new library is optional for existing applications, I think its useful to start new folks off on the path we'd like to go down in the future.

Refs cakephp/cakephp#6960

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1,0 +1,49 @@
+<?php
+namespace App;
+
+use Cake\Http\BaseApplication;
+use Cake\Error\Middleware\ErrorHandlerMiddleware;
+use Cake\Routing\Middleware\AssetMiddleware;
+use Cake\Routing\Middleware\RoutingMiddleware;
+
+/**
+ * Application setup class.
+ *
+ * This defines the bootstrapping logic, and middleware layers you
+ * want to use in your application.
+ */
+class Application extends BaseApplication
+{
+    /**
+     * Load all the application configuration and bootstrap logic.
+     *
+     * You can include any other files your application needs to bootstrap
+     * in a web context here.
+     *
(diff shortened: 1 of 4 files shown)


src/Application.php
```diff
@@ -0,0 +1,49 @@
+<?php
+namespace App;
```
ravage84: This file is missing the file header doc block.

webroot/index.php
```diff
@@ -24,14 +24,14 @@
         return false;
     }
 }
-require dirname(__DIR__) . '/config/bootstrap.php';
+require dirname(__DIR__) . '/vendor/autoload.php';
```
ravage84: Hm... People could wonder where or even if bootstrap is now called.

May be add a note down where ``Application`` gets instantiated?

markstory: Its in the Application code though.

src/Application.php
markstory: Thanks :+1:

lorenzo:
Great Job!

webroot/index.php
ionas: I tried to get 3.x installed through composer create-project (how?, could it be part of the regular readme?) to find my way around how/where Application.php is being executed because as @ravage84 points out it is not clear how/when/where boostrap.php is being loaded now.

markstory: Getting this code running with `composer create-project` will be tricky right now. Instead I would:

```
git clone git://github.com/cakephp/app.git appskel
cd appskel
git checkout -t -b 3.3 origin/3.3
composer install
```

PR #394, opened 2016-07-11
Change homepage header color to match our new theme.

ADmad:
Also reduced header height to show more useful info in initial viewport.


lorenzo:
Screenshot please :)

ADmad:
Before
app-before

After
app-after

Issue #395, opened 2016-07-11
Create a test case for the Application class


sdustinh:
I apologize if this has been covered somewhere else but I couldn't find it anywhere. With the 3.3 release in beta, I would love to see a sample `App\Test\ApplicationTest` added to the skeleton.

markstory:
Fixed now.

PR #396, opened 2016-07-11
Remove Application::bootstrap() as parent class has identical code.

ADmad:


src/Application.php
```diff
@@ -28,19 +28,6 @@
...
-     * Load all the application configuration and bootstrap logic.
-     *
-     * You can include any other files your application needs to bootstrap
-     * in a web context here.
-     *
-     * @return void
-     */
-    public function bootstrap()
-    {
-        require_once $this->configDir . '/bootstrap.php';
```
markstory: This was on purpose. My hope was that it made how the framework was working more obvious. Perhaps we could remove the base class method?

ADmad: I don't think changing how bootstraping is done would be commonly needed to warrant having this method in Application class. We could document in manual how bootstraping is done with code example showing how to override this method.

markstory: ok.

ADmad:
I also added test for `Application` as suggested in #395.

PR #397, opened 2016-07-11
Don't hide deprecation errors by default.

ADmad:


markstory:
Seems reasonable given that we've removed most of the runtime errors for the time being.

PR #399, opened 2016-07-15
apply new CS rule

antograssiot:


src/Console/Installer.php
```diff
@@ -179,12 +179,14 @@ public static function setSecuritySalt($dir, $io)
 
         if ($count == 0) {
             $io->write('No Security.salt placeholder to replace.');
```
ravage84: I don't think this improves readability, as mentioned [here](https://github.com/cakephp/cakephp/pull/9114#issuecomment-232915971)

config/bootstrap.php
```diff
@@ -162,10 +162,12 @@
  */
 Request::addDetector('mobile', function ($request) {
     $detector = new \Detection\MobileDetect();
+
     return $detector->isMobile();
 });
 Request::addDetector('tablet', function ($request) {
```
dereuromark: Instead of inside, there should more importantly be a newline after each closure..
Now it looks like the return belongs to the next statement etc more than it belongs to the actual own closure.

src/Console/Installer.php
jippi: seem like a 1 and 2 line block should just be without newline? :)

PR #401, opened 2016-07-19
Add fix for the httpoxy vulnerability

ravage84:
Refs https://github.com/cakephp/cakephp/issues/9137#issuecomment-233637635

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,3 +1,9 @@
+# Uncomment the following to prevent the httpoxy vulnerability
+# See: https://httpoxy.org/
+#<IfModule mod_headers.c>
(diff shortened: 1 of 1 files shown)


ADmad:
This should be added to `webroot/.htaccess` as for production it's recommended to set server's document root to `webroot` and thus this `.htaccess` won't be hit.

dereuromark:
Maybe we can also add a note to this htaccess that this recommended to use for local dev only.

jippi:
i think it should be `on` by default, 99,999% of the users wouldn't know it, but would out of the box be protected :)

markstory:
More pull requests are always welcome :smile:

PR #404, opened 2016-07-24
Use fluent interface for DRYer code.

ADmad:


src/Application.php
```diff
@@ -35,15 +35,16 @@ class Application extends BaseApplication
      */
     public function middleware($middleware)
     {
-        // Catch any exceptions in the lower layers,
-        // and make an error page/response
-        $middleware->push(new ErrorHandlerMiddleware());
+        $middleware
+            // Catch any exceptions in the lower layers,
+            // and make an error page/response
+            ->push(new ErrorHandlerMiddleware();
```
jippi: missing `)` and remove `;` ?

ADmad: At least we found out travis doesn't run the testsuite :P

antograssiot: @ADmad I wondered last time why you added the Application tests and I thought that it was to help users to write their own when adding Middleware.

ADmad: Since there was a `tests` folder with fixture and test cases I assumed there were valid tests for the files included in the skeleton app. I see now that to even make the testsuite run will require changes in `tests/bootstrap.php`.  Also besides the `PagesControllerTest.php` and `ApplicationTest.php` (which I have actually added in wrong directory) I really don't see the point of all the other files.

ADmad: Doh, the extra test files were created by me doing some baking in folder :P  I'll get the testsuite running on travis.

ADmad:
This is good to go. I updated travis config to run the test suite.

PR #417, opened 2016-08-16
Fix a small typo

BassemN:


inoas:
LGTM.

bravo-kernel:
Nice 🎣

config/bootstrap.php
```diff
@@ -15,7 +15,7 @@
 
 // You can remove this if you are confident that your PHP version is sufficient.
 if (version_compare(PHP_VERSION, '5.5.9') < 0) {
-    trigger_error('You PHP version must be equal or higher than 5.5.9 to use CakePHP.', E_USER_ERROR);
+    trigger_error('Your PHP version must be equal or higher than 5.5.9 to use CakePHP.', E_USER_ERROR);
```
dereuromark: Ideally, the version would be variableized :)

inoas: That is true, however we should do that as a global possibly. One that gets defined in index.php?

dereuromark: Inside each file at least one version would also be a good start probably.

inoas: A lot of boilerplate. Why not `define('CAKEPHP_PHP_MIN_VERSION', '5.5.9')`;
Another way would be to read it out from composer.json within index.php/bootstrap?
The book also references the minimum version, as do some tool setup files. Would be great to have those things centralised somewhere.

PR #418, opened 2016-08-16
Add markstory/asset_compress to composer suggest.

ADmad:


dereuromark:
:+1:

markstory:
I like this too :smile:

ADmad:
@markstory I am goin to ignore your bias here :P

PR #419, opened 2016-08-19
Update only to core bugfix versions.

ADmad:
Before this change the core would be updated to new minor version release too
but based on past experience it rarely goes smoothly and causes grief to users.
So only updating to bugfix versions is a better default.


markstory:
:+1: This will be safer for folks going forward.

PR #420, opened 2016-08-20
bump to php-56

antograssiot:


src/Controller/PagesController.php
```diff
@@ -31,14 +31,13 @@ class PagesController extends AppController
     /**
      * Displays a view
      *
+     * @param array ...$path
```
ADmad: I am not sure if `array` is the right type to use. Variadic args will always be an array so it would be more useful to hint the type of individual items in the array, `string` in this case. We could refer to what others do.

e.g. https://philsturgeon.uk/php/2013/08/30/potential-variadic-function-syntax-for-php-56/#pro-type-hinting

antograssiot: That was my that was my first intention but PHPStorm complained about it so I updated but I agree with you that string will better fit.
I'll update that as well as the travis build matrix where I forgot to remove 5.5

antograssiot:
@ADmad looks like we'll have to update cakephp/cakephp-codesniffer to accept variadics args 😞 
Should I revert this change fro now ?

src/Controller/PagesController.php
```diff
@@ -31,14 +31,13 @@ class PagesController extends AppController
     /**
      * Displays a view
      *
+     * @param string ...$path
```
ADmad: https://travis-ci.org/cakephp/app/jobs/153776773#L370

markstory:
Looking good so far.

antograssiot:
thank's @ADmad  I'am stupid

ADmad:
@antograssiot We all are at times :)

PR #422, opened 2016-08-21
Make code comments consistent

inoas:
- `/* */` for docs.
- `//` for disabled features

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -81,14 +81,14 @@ try {
     exit($e->getMessage() . "\n");
 }
 
-/*
+/**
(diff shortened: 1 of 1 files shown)


config/bootstrap.php
```diff
@@ -13,17 +13,23 @@
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 
-// You can remove this if you are confident that your PHP version is sufficient.
+/**
```
ADmad: `/**` is docblock standard and these comments aren't for API docblocks so IMO it's better to use `/*` instead.

PR #425, opened 2016-08-27
missing  onclick="this.classList.add('hidden');"

inoas:


markstory:
Thanks :+1:

ADmad:
Let's not add more inline js especially where it's not totally necessary.

inoas:
@ADmad I am totally with you. All it does is bringing it in line with the rest of the flash message default app templates.

ADmad:
I didn't know bake generated templates have this too :(

inoas:
They have delete confirm js, nothing else.

PR #427, opened 2016-08-27
Allow for HTML context.

dereuromark:
The default templates should allow for HTML context, e.g.
```php
$this->Flash->info('<b>' . h($highlight) . '</b> ' . h($message), ['params' => ['escape' => false]]);
```

### DIFF
--- a/src/Template/Element/Flash/success.ctp
+++ b/src/Template/Element/Flash/success.ctp
@@ -1,1 +1,6 @@
-<div class="message success" onclick="this.classList.add('hidden')"><?= h($message) ?></div>
+<?php
+if (!isset($params['escape']) || $params['escape'] !== false) {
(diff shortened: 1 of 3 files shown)


src/Template/Element/Flash/success.ctp
```diff
@@ -1 +1,6 @@
-<div class="message success" onclick="this.classList.add('hidden')"><?= h($message) ?></div>
+<?php
+if (!isset($params['escape']) || $params['escape'] !== false) {
```
markstory: Would this be simpler as:

```php
$params += ['escape' => true];
if ($params['escape']) {
  $message = h($message);
}
```

dereuromark: Sure that would also work.

PR #433, opened 2016-09-06
Add empty afterFilter() to App ErrorController

inoas:
Ref: https://github.com/cakephp/app/pull/432#issuecomment-244806693

### DIFF
--- a/src/Controller/ErrorController.php
+++ b/src/Controller/ErrorController.php
@@ -37,7 +37,7 @@ class ErrorController extends AppController
      * beforeFilter callback.
      *
      * @param \Cake\Event\Event $event Event.
-     * @return void
+     * @return \Cake\Network\Response|null|void
      */
     public function beforeFilter(Event $event)
     {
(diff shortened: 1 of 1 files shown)


src/Controller/ErrorController.php
```diff
@@ -55,4 +55,14 @@ public function beforeRender(Event $event)
 
         $this->viewBuilder()->templatePath('Error');
     }
+
+    /**
+     * afterFilter callback.
+     *
+     * @param \Cake\Event\Event $event Event.
+     * @return void
```
dereuromark: I am not so sure about the void part, especially for the other methods here.
If they do (and they can afaik) return a response, e.g. in beforeFilter() execution is bailing early.
IMO "Response|null" would be more correct for callbacks, right?

inoas: I can fix that up once there is an agreement on if to add it at all.

dereuromark: Yeah, I agree.
The problem with invalid voids is that the usage/check on those methods become yellow in IDEs like PHPStorm (for a good reason). As you should never have to save the return value of such a method into a variable or use it in the code.
We are doing that for these callbacks though inside the core code.
That clearly points out that those cannot be void by design.

dereuromark:
I am ok with this.

The return doc block line can be fixed to

    \Cake\Network\Response|null|void

to be consistent.

PR #434, opened 2016-09-06
Fix up doc block types. Pickyback class usage fix.

dereuromark:
Refs https://github.com/cakephp/app/pull/433#discussion_r77603622

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -49,7 +49,7 @@ class AppController extends Controller
      * Before render callback.
      *
      * @param \Cake\Event\Event $event The beforeRender event.
-     * @return \Cake\Network\Response|null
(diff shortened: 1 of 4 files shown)


src/Controller/AppController.php
```diff
@@ -49,7 +49,7 @@ public function initialize()
      * Before render callback.
      *
      * @param \Cake\Event\Event $event The beforeRender event.
-     * @return void
+     * @return \Cake\Network\Response|null
```
ADmad: phpcs doesn't like this https://travis-ci.org/cakephp/app/jobs/157901433#L360

dereuromark: Ah, needs explicit return null then.

dereuromark:
Not ideal, but it works.

PR #435, opened 2016-09-13
Use sudo false here by default.

dereuromark:
See https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments


lorenzo:
Thanks

PR #438, opened 2016-09-27
Move app checks before composer autoload

gonzela2006:


ravage84:
What is your reasoning behind it?
If I'm not mistaken, that would mean that these checks would either not be executed when using CLI, or would need to be copied in the shell script, too.

BassemN:
If you tried to run your app on PHP5.4 you will get the this error message
```Fatal error: Class 'DateTimeImmutable' not found in vendor/cakephp/chronos/src/Chronos.php on line 50```
so we need to make all checks before calling `require dirname(__DIR__) . '/vendor/autoload.php';` just like version app version 3.2 https://github.com/cakephp/app/blob/3.2.6/config/bootstrap.php#L16-L38

ravage84:
Well, you should not run your app using PHP 5.4, but with PHP 5.5.9+ instead:
https://github.com/cakephp/app/blob/master/composer.json#L8

ravage84:
As of CakePHP 3.2, the minimum PHP version is 5.5.
http://book.cakephp.org/3.0/en/appendices/3-2-migration-guide.html#minimum-php-5-5-required

BassemN:
Yes, I know so we need to run app checks earlier before calling `require dirname(__DIR__) . '/vendor/autoload.php';`

ravage84:
Your statement:

> If you run your app on PHP 5.4 you will get the this error message

Sounds like you have that problem on a machine with PHP 5.4.
PHP 5.4 is not supported anymore. Please try it on a machine with PHP 5.5.
If you still have the problem, we can further investigate.

BassemN:
I'm already running on PHP 5.5. I just found that we should add app checks before calling any code that depends on these checks

ravage84:
I see now.
But, as I said, your change means that at least two of those checks won't be executed when called trhough CLI.

The check for intl and mbstring would need to be added before https://github.com/cakephp/app/blob/3.3.2/bin/cake.php#L31, too.

BassemN:
In this case, we can add only PHP version check into `index.php` and leave the other checks into `bootstrap.php` I don't know if there is another better solution.

markstory:
If you're on PHP 5.5 how did you get the `DateTimeImmutable` missing error?

BassemN:
@markstory No, My app is running on PHP 5.5 but by mistake, I moved it into server with PHP 5.4 so I got the `DateTimeImmutable` missing error, yes, I know that CaskePHP requires PHP 5.5+ so the expected behaviour is to trigger this error https://github.com/cakephp/app/blob/3.3.2/config/bootstrap.php#L19-L21 and to trigger it it should be moved to `index.php` before the composer autoload

BassemN:
@markstory This doesn't happen in app 3.2x series because the auto was called into the `bootstrap.php` after all the checks as you can see here https://github.com/cakephp/app/blob/3.2.6/config/bootstrap.php#L17-L38

inoas:
Maybe we can abstract things away into a requirements_checks.php or so that is required by index.php. I'd love index php to just be an entrance point that basically almost never changes with cakephp/cakephp and/or cakephp/app changes and hoped to even see even the php webserver check moved into some env file within app/Config.

dereuromark:
I think it has to be a separate precheck.php file or something that is required by both index.php and the basic CLI process.

dereuromark:
Does anyone want to look into this for 3.5?

inoas:
Taking a look..

inoas:
I suppose #533 supersedes this...

dereuromark:
Alternative PR is now open.

Issue #440, opened 2016-10-10
IDs for base layout too generic


dereuromark:
I knew this was gonna haunt us some day.

When trying to bake some table with a "content" textarea field it gets a "width: 95" because `#content` apparently is used then for both base layout container and the textarea field itself.

So either we 
- reapply model prefixing for form fields (which I was always a big fan off to avoid those clashes in the first place)
- or we are less generic with the default classes and ids used in the main layout, e.g. `layout-` prefix: `#layout-content: {...}`

Our base template shipped with the app should be usable without lots of clashes, it should have good and sane default values to work with out of the box.

dereuromark:
Actually I don't see any usage of that id in the template right now.
Is that maybe a relict we can remove?
```
#content {
  margin:0 auto;
  padding-top:2em;
  width:95%;
}
```

inoas:
Just use `<main>` instead.

ADmad:
I don't think `<main>`  is constantly supported by browsers yet. Definitely not by IE.

IDs shouldn't be used for CSS anyway. Just switch to class and remove the ID.

dereuromark:
The error templates seem to be the only ones not fixed up to use classes yet. Reopening for now.

PR #441, opened 2016-10-10
Remove ID relict that breaks user land apps.

dereuromark:
Resolves https://github.com/cakephp/app/issues/440

### DIFF
--- a/webroot/css/cake.css
+++ b/webroot/css/cake.css
@@ -311,12 +311,6 @@ pre {
     color: #fff;
 }
 
-#content {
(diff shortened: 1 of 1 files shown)


ravage84:
Not sure if this is relevant:
https://github.com/cakephp/app/blob/9403b6b2c1205cb862619a7530b7b5052c080f6d/src/Template/Layout/error.ctp#L37

dereuromark:
Oh, we definitly have to change the error templates here to use classes then beforehand.

PR #443, opened 2016-10-27
Add composer scripts to the default app skeleton.

markstory:
Add some helpful composer script shortcuts for tasks people frequently do.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "license": "MIT",
     "require": {
         "php": ">=5.5.9",
-        "cakephp/cakephp": "3.3.*",
+        "cakephp/cakephp": "~3.3",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -6,7 +6,7 @@
     "license": "MIT",
     "require": {
         "php": ">=5.5.9",
-        "cakephp/cakephp": "3.3.*",
+        "cakephp/cakephp": "~3.3",
```
dereuromark: Most might find the ~ rather confusing I imagine, or confuse it maybe with ^.

ADmad: Using `~3.3` would auto update to 3.4 when it's released and based on past experience I am :-1: on auto updating to  next minor version.

markstory: Whoops, that's a mistake.

Issue #444, opened 2016-10-28
RFC - Add SecurityComponent by default 


steinkel:
suggested by Tim Summers

This is a (multiple allowed):
- [ ] bug
- [ ] enhancement
- [x] feature-discussion (RFC)
### Expected Behavior

SecurityComponent added to the application skeleton by default

In our latest CakePHP 3 training session, one of the attendees (Tim Summers) suggested this change to the app skeleton, and I think this could be a good idea to ensure people starting with the framework who didn't ever noticed the SecurityComponent exists would use it by default. It could be removed later on if needed.

What do you think?

markstory:
My biggest concern with enabling SecurityComponent by default is that it makes building forms that use Javascript harder to work with. Enabling `CsrfComponent` by default has similar challenges around AJAX forms.

dereuromark:
I am 👎 here, it should be enough to point out things in the docs.
But most apps might also be backend apps and there it is also often times counter-productive to enable this by default.

thinkingmedia:
We could add the component and then comment it out with a message explaining when you should use it?

I think @steinkel wants to promote it's usage for those who *should* be using it, but are not. It's not about forcing it upon users.

dereuromark:
Commenting out could be a compromise.

steinkel:
Closed per #446

PR #446, opened 2016-11-02
refs #444 add comment about security/csrf component

steinkel:
Promote use of Security/Csrf components

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -44,9 +44,9 @@ class AppController extends Controller
         $this->loadComponent('RequestHandler');
         $this->loadComponent('Flash');
 
-        /* 
-         * Enable the following components for recommended CakePHP security settings. 
-         * see http://book.cakephp.org/3.0/en/controllers/components/security.html 
+        /*
+         * Enable the following components for recommended CakePHP security settings.
+         * see http://book.cakephp.org/3.0/en/controllers/components/security.html
          */
         //$this->loadComponent('Security');
         //$this->loadComponent('Csrf');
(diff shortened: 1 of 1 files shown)


src/Controller/AppController.php
```diff
@@ -43,6 +43,10 @@ public function initialize()
 
         $this->loadComponent('RequestHandler');
         $this->loadComponent('Flash');
+        //Recommended security enhancements for your application
```
ravage84: For better readability, may be add a white line.

Please add a space affter ``//`` for comments.
For the commented out code, it's OK.

dereuromark: I agree:

    // See http...

inoas: In here it is different: https://github.com/cakephp/app/blob/master/config/bootstrap.php#L202-L205

Maybe /* Recommended security enhancements for your application */ ?

dereuromark: @ionas Those should only be used for multiline comments IMO

inoas: Don't agree, neither does this https://github.com/cakephp/app/blob/master/config/bootstrap.php#L168-L170

I'd say `/* foo */` is for inline comments whereas `//$code` would be for out-commented lines.

bravo-kernel:
Instructions seems/felt interpretable to me. Maybe:

Uncomment settings to enable CakePHP recommended security settings. Or sth similar

inoas:
`/* Enable following components for recommended CakePHP security settings. */`

bravo-kernel:
👍🏼 Glad you could make some sense out of my drive-and-comment jibberish

PR #450, opened 2016-11-19
Allow IDE typehinting/clickability

dereuromark:


config/routes.php
```diff
@@ -44,6 +44,8 @@
 Router::defaultRouteClass(DashedRoute::class);
 
 Router::scope('/', function (RouteBuilder $routes) {
+    /* @var \Cake\Routing\RouteBuilder $routes */
```
lorenzo: WAT

dereuromark: What class should it be?

dereuromark:
Ah never mind. I copy and pasted from my app Code. Seems like the typehint is now already finally included :) im the Funktion Signature
Didn't see that.

config/routes.php
lorenzo: No, I just find it funny that the variable is type hinted but you sill need to add another hint in th next line

dereuromark: I just had outdated app Code ;) and it was late.

Issue #452, opened 2016-11-25
Migrations


inoas:
Should Plugin::load('Migrations'); not be in bootstrap_cli together with bake?

markstory:
It could. The webserver part of an app normally doesn't need migrations.

inoas:
Closing as I quickly did the changes via GH

BassemN:
Loading `Migrations` plugin only in CLI breaks migrations in [a non-shell environment](http://book.cakephp.org/3.0/en/migrations.html#running-migrations-in-a-non-shell-environment).

PR #453, opened 2016-11-25
Remove Migrations Plugin from bootstrap.php

inoas:
Ref: https://github.com/cakephp/app/issues/452
Note: 2 PRs cause quick via GH.


inoas:
This should also help a tiny fraction with speed <3 :)

dereuromark:
Sorry, but it does sure not :) Speed is irrelevant for a simple plugin load call like this.

PR #454, opened 2016-11-25
Add Migrations Plugin to bootrap_cli.php

inoas:
Ref: https://github.com/cakephp/app/issues/452
Note: 2 PRs cause quick via GH.

While I like it to be optional, should it be outside the try block instead?


dereuromark:
It should definitely be outside the try catch, as it should always be present.

dereuromark:
Should we add it to the end?
Then "plugin load .. -cli" calls could easily write their plugins to the end after that one. Refs https://github.com/cakephp/cakephp/issues/9150

PR #455, opened 2016-11-27
Use ::class syntax where applicable

dereuromark:
Follows https://github.com/cakephp/cakephp/pull/9806

Must only be merged after 3.4 is stable.


markstory:
Could this be rebased on top of `3.next` if its a 3.4+ thing.

dereuromark:
Sry, was an oversight. Done.

PR #458, opened 2016-12-05
Bootstrap: HTTP_X_FORWARDED_PROTO

dorxy:
When the bootstrapper sets the fullBaseUrl, it now takes into account any protocols forwarded in the HTTP_X_FORWARDED_PROTO environment variable. This is very useful when applications are behind load balancers which forward SSL while the application itself runs without SSL.

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -140,7 +140,7 @@ if ($isCli) {
  */
 if (!Configure::read('App.fullBaseUrl')) {
     $s = null;
-    if (env('HTTPS')) {
+    if (env('HTTPS') || env('HTTP_X_FORWARDED_PROTO', 'http') == 'https') {
         $s = 's';
     }
 
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -140,7 +140,7 @@ if ($isCli) {
  */
 if (!Configure::read('App.fullBaseUrl')) {
     $s = null;
-    if (env('HTTPS') || env('HTTP_X_FORWARDED_PROTO', 'http') == 'https') {
+    if (env('HTTPS') || env('HTTP_X_FORWARDED_PROTO', 'http') === 'https') {
         $s = 's';
     }
 


josegonzalez:
@chinpei215 I think you mentioned this exact change ;)

config/bootstrap.php
```diff
@@ -140,7 +140,7 @@
  */
 if (!Configure::read('App.fullBaseUrl')) {
     $s = null;
-    if (env('HTTPS')) {
+    if (env('HTTPS') || env('HTTP_X_FORWARDED_PROTO', 'http') == 'https') {
```
dereuromark: Please always use strict comparison with strings.

chinpei215:
@josegonzalez Yeah. But I think it should be commented out by default, as malicious users can send the header to cheat our applications.

dorxy:
Isn't that also the case with Controller::clientIp ? Why would that not be concidered malicious? Trying to learn more :)

ADmad:
@dorxy It is the same with `Request::clientIp()` too which is by `Request::$trustProxy` is false by default.

dorxy:
Ah I didnt know that was a parameter, thanks! I'll update it, perhaps we can use that same parameter for determining whether to use the protocol variable?

ADmad:
> perhaps we can use that same parameter for determining whether to use the protocol variable

No you can't since there is no request instance available yet when bootstrapping :)

dorxy:
Ok cool, I'll update the PR. Would you like it as a separate, within or 1 statement diff if?

chinpei215:
@dorxy By the way, we have another approach. We can do like the follwoing in .htaccess (I was taught this from @nojimage. Thanks):
```apache
SetEnvIf X-Forwarded-Port  443   HTTPS=on
SetEnvIf X-Forwarded-Proto https HTTPS=on
SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1
```
What do you think about this approach? I am not sure whether changing 'App.fullBaseUrl' is the best way, as I think `$this->rqeuest->is('ssl')` doesn't return `true`.

dereuromark:
Sounds like more a documentation issue then.

ADmad:
Closing as the suggested could open up security issues. A PR/issues can be opened on the docs repo if required.

Issue #442, opened 2016-10-22
Set productive setting as default always


dereuromark:
According to https://github.com/cakephp/app/blob/master/config/bootstrap.php#L96-L99
The productive cache lengths are set if !debug:

``` php
if (!Configure::read('debug')) {
    Configure::write('Cache._cake_model_.duration', '+1 years');
    Configure::write('Cache._cake_core_.duration', '+1 years');
}
```

But in in fact, similar to debug mode, it should always be set by default to production system and then optionally modified for local dev:

``` php
// Set production defaults in app.php to very long time (+1 year)

if (Configure::read('debug')) {
    Configure::write('Cache._cake_model_.duration', '+10 seconds');
    Configure::write('Cache._cake_core_.duration', '+10 seconds');
}
```

etc.

markstory:
Ok.

lorenzo:
Sounds like a good idea to me

ADmad:
@dereuromark Would you like to make a PR so that this issue can be closed?

dereuromark:
I'd appreciate it if someone had the time to make the PR.

dereuromark:
PR open.

PR #460, opened 2016-12-18
Set the default cache duration to 1 year

cleptric:
This implements #442

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -93,9 +93,9 @@ try {
  * for a very very long time, as we don't want
  * to refresh the cache while users are doing requests.
  */
-if (!Configure::read('debug')) {
-    Configure::write('Cache._cake_model_.duration', '+1 years');
-    Configure::write('Cache._cake_core_.duration', '+1 years');
+if (Configure::read('debug')) {
+    Configure::write('Cache._cake_model_.duration', '+2 minutes');
+    Configure::write('Cache._cake_core_.duration', '+2 minutes');
 }
 
 /*
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -89,9 +89,8 @@ try {
 //Configure::load('app_local', 'default');
 
 /*
- * When debug = false the metadata cache should last
- * for a very very long time, as we don't want
- * to refresh the cache while users are doing requests.
+ * When debug = true the metadata cache should only last
+ * for a short time.
  */
 if (Configure::read('debug')) {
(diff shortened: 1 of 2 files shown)


config/bootstrap.php
```diff
@@ -93,9 +93,9 @@
  * for a very very long time, as we don't want
  * to refresh the cache while users are doing requests.
  */
-if (!Configure::read('debug')) {
-    Configure::write('Cache._cake_model_.duration', '+1 years');
-    Configure::write('Cache._cake_core_.duration', '+1 years');
+if (Configure::read('debug')) {
+    Configure::write('Cache._cake_model_.duration', '+2 minutes');
+    Configure::write('Cache._cake_core_.duration', '+2 minutes');
```
dereuromark: Could we even go for 1 min?

dereuromark: Lets wait what others think.
So far the PR looks great!

markstory: 2 minutes has been fine so far as far as I know.

ADmad: Cake 2 has 10 seconds :)

dereuromark: Yeah, I would also rather shorten it a bit by default - and people can increase once they know "why" those files could be cached and changes in the DB are not immediately reflected.

inoas: I'd not expect cache to be on at all when debug is on. If there is a good reason to have it on, 10 seconds sounds sane oob.

dereuromark: Depending on the engine, they might not support to large time-frames.
And IMO 1 year is more than fine. A small re-cache for a single entity at this point in time is acceptable as default for the framework config here.
But the debug mode time should indeed be more like <= 1 minute again.
I think 3.x needs a bit more here since the DB caching of schema is quite time intensive afaik.

inoas: If "endless caching" is not an option, then I'd opt for 1 month, at that frame you can easily detect if there are issues alongside re-caching and are not suddenly hit by it "one year after deployment".

For debug caching, what would you suggest if 10 seconds ain't possible?

markstory: If you aren't deploying or modifying an app for 1 year. It seems unlikely that rebuilding a cache would be that harmful, as it is liekly 'unmaintained' code at that point.

cleptric: As this PR was mainly done because of https://github.com/cakephp/app/issues/442, I guess it fulfils his purpose of changing the default behavior of setting the cache settings.
The values are changeable at any time by the developer and worked well since cake 3 was release.

Issue #462, opened 2016-12-26
How to upgrade cakephp/app?


littleylv:
With my many thanks for your CakePHP.

I got ``composer update`` to update vendors in my application. But I'm wondering how to upgrade cakephp/app?

I found you guys make many wonderfull PRs to cakephp/app which could be very useful for everyone's application.

I created my app using:
``composer create-project --prefer-dist cakephp/app [app_name]``

ADmad:
You will have to do the updates manually. Since this skeleton is a starting point for your app and the various files would have been modified as per your needs there isn't a way to provide automated upgrade.

littleylv:
I got you. @ADmad

PR #470, opened 2017-01-05
Add comment on updating the app skeleton

ravage84:
Refs: https://github.com/cakephp/app/issues/462

@ADmad hope you don't mind that I borrowed your summary on the situation.


inoas:
Maybe point to app.default.php, bootstrap.php, email.php Application.php and index.php as these are changed more often?

ravage84:
I'd rather create a section in the CakePHP Cookbook for that and link from here to there.

PR #473, opened 2017-01-12
Add Type::build('timestamp')

gorogoroyasu:
added two lines to config/bootstrap.php, which will convert timestamp column data in Entity not into Cake\I18n\Time object but into Cake\I18n\FrozenTime object.
```
Type::build('timestamp')
    ->useImmutable();
```

## Description
When I was using PostgreSQL, the datetime column data in Entity Object was not converted to the object of `Cake\I18n\FrozenTime`, but was converted to the object of `Cake\I18n\Time`.

In the end, the behavior of application was not a bug, though I was really confused.
By inserting two lines I added, the timestamp will be automatically converted into `Cake\I18n\FrozenTime`.
I wish this change will help PostgreSQL or other DB users.

Even if you reject this PR, please consider a bit more to remain these two lines as commented out lines.

## Comment
At least, the document should show other DB users some cautions shown above, I think. So, nevertheless this PR is merged or rejected, I will create a PR to cakephp/docs.

I apologize you to skip the process below.
```
The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.
```

Thank you.

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -192,6 +192,8 @@ Type::build('date')
     ->useImmutable();
 Type::build('datetime')
     ->useImmutable();
+Type::build('timestamp')
+    ->useImmutable();
 
 /*
  * Custom Inflector rules, can be set to correctly pluralize or singularize


inoas:
Looks sane to me, however I wonder why Postgres Datetime seems to be recogonized as `timestamp` not `datetime` - smells bad to me?

gorogoroyasu:
@inoas 
Maybe, this line cause it.

https://github.com/cakephp/cakephp/blob/master/src/Database/Schema/PostgresSchema.php#L360

gorogoroyasu:
@dereuromark thanks for fixing the title.

dereuromark:
Maybe @lorenzo can answer that. I would also prefer fixing the core here and have a consistent usage rather than adding a non agnostic thing in the app template for a single DB type.

chinpei215:
@gorogoroyasu [This line](https://github.com/cakephp/cakephp/blob/a0ce4c453c53ba9e9ee5b29e58f854fa48e516ea/src/Database/Schema/PostgresSchema.php#L93) causes this issue. The line you are pointing to is used for generating `CREATE TABLE` statement.

@dereuromark This issue occurs on MySQL too, if tables contain `TIMESTAMP` type. If we map `TIMESTAMP`  of DB to `datetime` of Cake, the table would never be converted to `CREATE TABLE` statement correctly in MySQL when creating fixtures. Probably, that is the reason why we map `TIMESTAMP` of DB to `timestamp` of Cake.

dereuromark:
@chinpei215 That makes sense, but doesn't https://github.com/cakephp/cakephp/blob/master/src/Database/Schema/PostgresSchema.php#L360 also do DateTime => Timestamp mapping?

chinpei215:
@dereuromark If we map `TIMESTAMP` to `datetime` when introspecting field types, then we should map `datetime` to `TIMESTAMP` when generating CREATE TABLE statement, shouldn't we?

markstory:
Thank you :clap:

inoas:
So what about the change of mapping for the postgres adapter (aka datetime -> datetime instead of datetime -> timestamp)?

chinpei215:
@inoas Are you suggesting to map `TIMESTAMP WITHOUT TIME ZONE` of Postgres to `datetime` of Cake?

PR #474, opened 2017-01-14
Allowing beta versions of libs is a bad default.

ADmad:


markstory:
Thanks! This is probably why the 3.3.12 zipball was built wrong.

PR #475, opened 2017-01-14
New home page design

andrej-griniuk:
New home page design


bravo-kernel:
Screenshot?

andrej-griniuk:
screen

bravo-kernel:
Sweet 😋

bravo-kernel:
Probably has nothing to with your theme but I just noticed Slack is not in the help list.

src/Template/Pages/home.ctp
```diff
@@ -35,214 +35,224 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
     <title>
         <?= $cakeDescription ?>
     </title>
+
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
```
antograssiot: I would rather don't use any external URL. 
It will probably looks ugly when working offline.

Also we already have foundation, seems odd to add bootstrap

markstory: > Also we already have foundation, seems odd to add bootstrap

:+1: for not having 2 CSS libraries in use.

ADmad:
Looks nice, but bootstrap needs to be removed.

andtxr:
I think that nowadays Bootstrap is fairly more popular than Foundation. Maybe change the layout base of the CakePHP app should be something to debate in a near future. Even the main CakePHP website now uses Bootstrap.

bravo-kernel:
😳

dereuromark:
Maybe we can completely move to bootstrap - that would solve two issues at once.

andrej-griniuk:
+1 for completely moving to Bootstrap. It's kind of industry standard nowadays.

inoas:
Looks nice but we should remove any external dependencies. Example: If you are on a train with your notebook you don't want to load those external ressources when developing an application locally.

dereuromark:
I am with admad, Just foundation is fine.

inoas:
For 3.x it is for sure as that has been chosen as the basis.

ADmad:
You would be seeing this homepage only after initial install for which you need internet access anyway for composer to work. After that no one cares about this page. So external dependency is fine.

A new PR is already updating this template to use foundation instead of bootstrap.

PR #478, opened 2017-01-15
New home page design reworked using Foundation

andrej-griniuk:


antograssiot:
Thanks for your work. :+1:

src/Template/Pages/home.ctp
```diff
@@ -35,224 +35,242 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
     <title>
         <?= $cakeDescription ?>
     </title>
-
-    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
     <?= $this->Html->meta('icon') ?>
+    <?= $this->Html->css('base.css') ?>
+    <?= $this->Html->css('cake.css') ?>
     <?= $this->Html->css('home.css') ?>
-    <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.6.3/css/font-awesome.min.css">
```
antograssiot: So FA was not needed/used after all ?

cleptric:
@andrej-griniuk Could you also add the header image for @2 and @3 pixel-ratios? Looks a little bit pixelated on my screen :)

inoas:
As @antograssiot already noted, could external dependencies be removed, just copy the libraries as they are MIT it is not a problem. External dependencies create problems in intranets, are bad for privacy and good for unwanted analysis by third parties and make mobile internet connections slow. They also make the projects initial home page appear to render more slowly as they are render blocking (above the fold blocking).

webroot/css/home.css
```diff
@@ -59,46 +26,115 @@ a {
...
+    font-family: 'Roboto', sans-serif;
+    font-size: 27px;
+    line-height: 30px;
+    font-weight: 300;
+    -webkit-font-smoothing: antialiased;
+    margin-top: 0;
+    margin-bottom: 20px;
+}
+
+.home h3.more {
```
markstory: Do you need this much specificity or would `.home .more` or even `.more` work?

webroot/css/home.css
```diff
@@ -59,46 +26,115 @@ a {
...
+    color: #ffffff;
+    background-color: #d33d44;
+    padding: 15px;
+    margin-top: 10px;
+}
+
+.home header .header-title {
```
markstory: Do you need this much specificity or would `.home .header-title` work?

webroot/css/home.css
```diff
@@ -59,46 +26,115 @@ a {
...
+    padding: 0;
+    display: block;
+    background: #404041;
+    text-align: center;
+}
+
+.home header .header-title h1 {
```
markstory: Do you need this much specificity or would `.header-title h1`?

webroot/css/home.css
```diff
@@ -59,46 +26,115 @@ a {
...
 }
 
-.alert-warning p::before {
+.home .alert p {
+    margin: 0;
+    font-size: 12px;
+    line-height: 1.4;
+}
+
+.home .alert p:before {
```
markstory: Do you need this much specificity?

webroot/css/home.css
```diff
@@ -108,19 +144,36 @@ a:hover {
...
-li {
+.home li {
     padding-left: 1.8em;
 }
 
-li.success:before {
-    color: #88c671;
-    content: "\0056";
-    font-family: 'cakefont';
+.home li.bullet:before {
```
markstory: Would `.bullet:before` work?

webroot/css/home.css
```diff
@@ -130,119 +183,63 @@ li.success:before {
...
     font-size: 60px;
 }
-
-.docs {
+.home .icon.docs {
     font-size: 57px;
 }
-
-.training {
+.home .icon.training {
```
markstory: Would `.icon.training` work?

src/Template/Pages/home.ctp
andrej-griniuk: Nope.

andrej-griniuk:
@inoas so the only external dependency left is fonts. Don't think it's worth distributing them with the app. They're used on a page most of the users will only see once. If worse comes to worst - they'll see it with fallback font.

andrej-griniuk:
@cleptric replaces it with SVG.

lorenzo:
@andrej-griniuk would you mind attaching a screenshot?

andrej-griniuk:
screen

src/Template/Pages/home.ctp
```diff
@@ -35,224 +35,242 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
...
-                <?php Debugger::checkSecurityKeys(); ?>
-                <div id="url-rewriting-warning" class="alert alert-warning text-center url-rewriting">
-                    <p>
+<header class="row">
+    <div class="header-image"><?= $this->Html->image('cake.logo.svg') ?></div>
+    <div class="header-title">
+        <h1>Welcome to CakePHP <?= Configure::version() ?> Red Velvet. Build fast. Grow solid.</h1>
+    </div>
+</header>
+<div id="content">
```
dereuromark: Do we need id selectors here?

lorenzo:
Great job @andrej-griniuk

inoas:
@andrej-griniuk so when you are on a low brand-with connection the home page will slowly load on some browsers because it is neither stored alongside the app skeleton nor loaded below-the-fold.

In head
```css
<link href="https://fonts.googleapis.com/css?family=Raleway:500i|Roboto:300,400,700|Roboto+Mono" rel="stylesheet">
```

lorenzo:
That's a fair concern... should we not rely on external fonts or maybe load them at the bottom?

cleptric:
We can use Arial to be super save :)

inoas:
Above the fold you can just use, and it will pick up whatever it can find:

```css
font-family: Raleway, "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
```

Then you need to load Raleway non-blocking.

lorenzo:
@ionas do you think you could create a pull request with that change?

PR #479, opened 2017-01-15
Use __d() function to allow to translate

chinpei215:
Link to the official Slack team
Link to SSL pages if possible

... but I would have to resolve conflicts after #478 had been merged 😞


dereuromark:
Could it also be "app" instead of "cake_app"? Or is the latter necessary to clarify that this is a CakePHP app sleketon?

chinpei215:
> Could it also be "app" instead of "cake_app"?

It could. If there are no objections, I'll do that when I resolve conflicts.

chinpei215:
Closing. I'll re-send another pull request when I have time.

PR #480, opened 2017-01-15
Update the favicon

cleptric:
We should use the same favicon as cakephp.org to be more consistent regarding the CakePHP CD


inoas:
An SVG and PNG variant would be cool:
http://caniuse.com/#feat=link-icon-svg (the future)
http://caniuse.com/#feat=link-icon-png (the current)
ico = the past

markstory:
Looks good to me :+1:

PR #483, opened 2017-01-16
Defer Font Loading

inoas:
Ref: #478
I am sure there are other and better ways to do it. Feel free to improve.

### DIFF
--- a/src/Template/Pages/home.ctp
+++ b/src/Template/Pages/home.ctp
@@ -39,7 +39,6 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
     <?= $this->Html->css('base.css') ?>
     <?= $this->Html->css('cake.css') ?>
     <?= $this->Html->css('home.css') ?>
-    <link href="https://fonts.googleapis.com/css?family=Raleway:500i|Roboto:300,400,700|Roboto+Mono" rel="stylesheet">
 </head>
 <body class="home">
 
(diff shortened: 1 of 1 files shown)


markstory:
Won't this result in unstyled text, and a reflow when the fonts are loaded?

phpnut:
What is the purpose of this?

inoas:
@phpnut See: https://github.com/cakephp/app/pull/478#issuecomment-272883579

@markstory Yes and no:
- The font stack should have fallback fonts that look "good enough" and render with a similar x-height and average letter-width, this helps: https://meowni.ca/font-style-matcher/
- People can have installed the font locally

If the fonts block rendering on slow connections you will have a white home page.

I suspect the next iteration of the app layout / bake will also feature brand-red colors and the Raleway font, so the problem will persist throughtout cakephp/app and cakephp/bake. Then the loading issue will be on every baked request.

Alternative: publish the font alongside cakephp/app - but this was dismissed without giving a reason.

markstory:
> Alternative: publish the font alongside cakephp/app - but this was dismissed without giving a reason.

Is that even possible? Are the licenses on the fonts compatible with MIT?

lorenzo:
I'd rather have a page reflow than a page with no text while waiting on a slow connection.

markstory:
@lorenzo That's fair.

lorenzo:
@ionas the using it, do you notice any flashing of content?

inoas:
After 2-3 hours of debugging I can say that FOIT/FOUT is not a problem, but I cannot for 100% confirm that the javascript in this PR solves the blocking problem. It is hard to make just the external network slow, inspector throttling makes the whole connection slow and is missing in firefox (only available in chrome).

So I am going to close this. If people have real problems with loading the homepage on sluggy connections, feel free to reopen/open a new one.

inoas:
@markstory @phpnut as for if we can bundle a SIL Open Font Licensed font see https://github.com/FortAwesome/Font-Awesome/issues/1124#issuecomment-27427244

Bundling would solve it in a good way.

Issue #481, opened 2017-01-15
RFC: Update default css


cleptric:
As CakePHP has a pretty neat design since 3.3, I'm always a little bit disappointed to see this "not-so-cool" looking design when I bake my templates.
Can we update the css and use more of the "cake-red" instead of the green?
Maybe some additional eye candy here and there ;)
If this is ok with you, I can give it a try.

markstory:
I don't know if @phpnut has any designs from the rebranding designer. If not switching green for red would be a great place to start.

inoas:
The bake templates should not rely on external font resources, IMHO. A local copy of Raleway or whatever it is and a swap on the color palette would do a lot 👍

cleptric:
As a PR is already open and additional a new layout will be provided in the future, I'll close this

PR #485, opened 2017-01-18
Adopt CakePHP CD for cake.css

cleptric:
I updated the cake.css file like proposed in #481
My "concept" was to simplify the basic color palette a bit.
The main color now is #D33C43, with a 10% darker and a 10% lighter accent alternative.
I also added the cake logo to the header.
I would call this a first step, because there are still some other things to improve.

Before:

screen shot 2017-01-18 at 21 54 13

Now:

screen shot 2017-01-18 at 23 14 48

EDIT: Now shows state of the last commit


josegonzalez:
The red links look super aggressive. Can we use the same colors from the book?

cleptric:
Sure. We could also style the body in different shades of grey (not kidding) and just use the red for the header.

webroot/css/cake.css
```diff
@@ -350,20 +350,33 @@ pre {
 }
 
 .top-bar.expanded .title-area {
-    background: #01545b;
+    background: #ba232a;
+}
+
+.top-bar.expanded .title-area:before {
```
markstory: Any way we could reduce the specificity here?

cleptric: `.title-area:before` is fine, too. 
Just out of curiosity, is less specific best practice?

phpnut:
I do have new designs coming for app/. I forgot to get them with the redesign. Don't get discouraged when your hard work here gets replaced. Thanks for you effort and it is looking good.

webroot/css/cake.css
```diff
@@ -553,4 +566,4 @@ div.message.hidden {
 table td {
     vertical-align: top;
     word-break: break-all;
-}
+}
```
dereuromark: As a side note: Make sure your IDE is properly configured (ending newline).

josegonzalez:
@cleptric The links on the side should be something other than black, to make them more "link-like" :)

cleptric:
@josegonzalez Use the colors of the book he said 😂 That are the same colors as the links in the sidebar of the book. But I'll change them anyway 😀

josegonzalez:
Oh nvm I'm a noob. Carry on then!

Oxicode:
> The red links look super aggressive

Its tue

josegonzalez:
@phpnut is this still waiting on new designs?

cleptric:
I'll close this for now. When we have the new design we can reopen it 🙂

PR #486, opened 2017-01-18
Add coverage composer script

dereuromark:
This is a super useful helper I use myself, and since we are now using composer scripts, here it goes:

    composer coverage

It will then write a full HTML coverage report to browse into /webroot/coverage/
Just browse afterwards to your

    http://example.local/coverage/index.html

and you can then click through the local coverage report generated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /config/app.php
 /tmp/*
 /logs/*
+/webroot/coverage/*
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,7 @@
         ],
         "cs-check": "phpcs --colors -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests",
         "cs-fix": "phpcbf --colors --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests",
-        "test": "phpunit --colors=always"
+        "test": "phpunit --colors=always",
+        "coverage": "phpunit --log-junit webroot/coverage/unitreport.xml --coverage-html webroot/coverage --coverage-clover webroot/coverage/coverage.xml"
     }
 }


josegonzalez:
Should we add that directory to gitignore as well?

chinpei215:
I am not in favor of this change. Why do you publish HTMLs under the webroot? If I were a hacker, I would try to access with `/coverage/` when I found a CakePHP3 app.

markstory:
@chinpei215 Do you typically run tests on your production hosts?

dereuromark:
This is for local Dev only if course. Check the docs, same tip ist already there as well.

bravo-kernel:
Very useful. Since I dump my coverage files outside my vm's webroot... would it be possible to use an argument for the directory? Maybe default to webroot if no directory argument is given.

chinpei215:
@markstory No, I don't.
@dereuromark If you run phpunit --coverage-html on local only, why do you need to publish it under the webroot? You can see any html files in your browser by using `file://` schema.

dereuromark:
Nothing can happen live because without require dev dependencies there is no phpunit :) and the folder is gitignore and will never be commited.

josegonzalez:
lets move it up a directory, maybe to `tmp`?

bravo-kernel:
I actually think `webroot` would be a good default for novice users since that will always be browseable andt thus "just work", even on a remote vm, without requiring any additional trickery. Simple instruction for the book as well IMO.

dereuromark:
Exactly what I was thinking :)

ADmad:
Remember back in 1.x days how many live sites had `webroot/test.php`?

bravo-kernel:
@ADmad fair point, would this not all be solved when using a directory argument (if technically possible)?

antograssiot:
I don't he buy the gitignore argument.
Do you really think that novice users will use git to deploy their app ?

dereuromark:
I am also fine with tmp dir ;) etc

ADmad:
@bravo-kernel A directory argument with defaulting to `TMP` should be fine.

dereuromark:
Someone wants to modify the PR?

bravo-kernel:
@ADmad that would be awesome, without an argument this would be useless to me. Nice addition @dereuromark

dereuromark:
I think we should rather get this going and merge, and improve it more over time. Discussing it dead is doing less here for all.

thinkingmedia:
@dereuromark Correct me if I'm wrong, but doesn't the coverage report use relative paths as to allow it to be opened from any directory location. For example; you could write the reports to `app/tmp/coverage` and just double click the `index.html` file to view the report. It doesn't need a web service or webroot.

dereuromark:
The idea is to allow local URLs to quickly be used here, e.g. when you set up your vhost via cakebox and my-app.local, you just browse to my-app.local/coverage.
It is super easy this way - and due to the fact of require-dev only happening/possible for local dev.

thinkingmedia:
@dereuromark okay, this looks good to me 👍 

If someone doesn't like it. They don't have to run it or they can change it.

bravo-kernel:
@dereuromark could be me but I thought the discussion was already over and we were just waiting for the requested change.

> add a --directory argument defaulting to tmp

dereuromark:
@bravo-kernel then please do that yourself afterwards.

bravo-kernel:
@dereuromark sorry but I think it will be best for me to leave this discussion. Could be the flu but I'm getting a very weird feeling about this PR, not even seeing the `tmp` change feels the opposite of :heart somehow.

dereuromark:
We just discussed that this is not comparable to any of the issues of 2.x days - the comparison is flawed.
There are no issues here.
And you yourself said, that it should be simple for newbies, and we all agree that it is easily modifyable if anyone wants to have it differently for their apps.

A follow up PR can still fine tune more, but that will need a whole lot of more discussion then and is out of scope of this simple addition here right now IMO.

dereuromark:
I do listen, and I do value @bravo-kernel s feedback.
But all that now happened was a perfectly fine PR to become stale.

You as commenting people also have to value and respect the work that the person doing the PR did.
And that also means that if the points/issues raised have been covered or talked about and the consensus then was that it is if no issue after all, then there should be no need to block this from moving forward.
Incremental changes are also way better than no progress at all.

bravo-kernel:
Apologies if I caused any grievance here, that was not my intention. I was under the impression there was consensus of not using webroot so (to me) the adjustment seemed logical and up to mr. @dereuromark. In that context I might easily have misinterpreted things as "thanks for the time taken to think about improving this but please go fix it yourself", my bad.

dereuromark:
I want each other to work together - not block each other from doing sth :) That's all.

dereuromark:
I will close this now. If someone wants to see it happening in this or a different form, please PR.

PR #493, opened 2017-01-29
Add coverage script

cleptric:
My try to implement #486 

I created a new Console class for that.
You now can simply run `composer coverage` and the generated files will go inside the `tmp` directory. If you want to change the path, you can type `composer coverage -- <custom-directory>`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,8 @@
         ],
         "cs-check": "phpcs --colors -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests",
         "cs-fix": "phpcbf --colors --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests",
-        "test": "phpunit --colors=always"
+        "test": "phpunit --colors=always",
+        "test-coverage": "App\\Console\\Coverage::generateCoverage"
     },
     "prefer-stable": true
 }
--- a/src/Console/Coverage.php
+++ b/src/Console/Coverage.php
@@ -1,0 +1,37 @@
+<?php
+namespace App\Console;
+
+use Composer\Script\Event;
+
+/**
+ * Generate phpunit coverage reports
+ */
+class Coverage
+{
+    /**
+     * Creates the phpunit coverage report
+     *
+     * Run `composer coverage -- <custom-directory>` to change the default output path
+     *
+     * @param \Composer\Script\Event $event The composer event object.
+     * @return void
+     */
+    public static function generateCoverage(Event $event)
+    {
+        $io = $event->getIO();
+        $rootDir = dirname(dirname(__DIR__));
+        $coverageDir = $rootDir . '/tmp/';
+
+        if (isset($event->getArguments()[0])) {
+            $coverageDir = $rootDir . '/' . $event->getArguments()[0];
+        }
+
+        exec('phpunit --log-junit ' . $coverageDir . '/coverage/unitreport.xml --coverage-html ' . $coverageDir . '/coverage --coverage-clover ' . $coverageDir . '/coverage/coverage.xml', $out, $status);
+
+        if ($status !== 0) {
+            $io->write("Command failed with status: $status \n");
+        } else {
+            $io->write("Coverage report succesfully created in $coverageDir");
+        }
+    }
+}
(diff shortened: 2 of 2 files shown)


src/Console/Coverage.php
```diff
@@ -0,0 +1,37 @@
+<?php
```
ravage84: Since this is neither a CakePHP shell nor a task, I find it somewhat confusing to find this file directly within the Console folder.
I see, there is already an [Installer.php](https://github.com/cakephp/app/blob/3.3.6/src/Console/Installer.php), which is also a Composer related script.

I would except to find such files within a sub folder named Composer, for example.

cleptric: REPOST: Same for `ConsoleShell` inside `src/Shell`. Should we move both into subfolders. Maybe in a different PR?

ravage84: @cleptric I see a a compatibility problem, here.

cleptric: But this would only affected newly created projects, right?

composer.json
```diff
@@ -42,6 +42,7 @@
         ],
         "cs-check": "phpcs --colors -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests",
         "cs-fix": "phpcbf --colors --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests",
-        "test": "phpunit --colors=always"
+        "test": "phpunit --colors=always",
+        "coverage": "App\\Console\\Coverage::generateCoverage"
```
dereuromark: should we name it

    "test-coverage": ...

to be more precise that this is coverage for tests?

ravage84: I'm OK with either.

dereuromark:
@cleptric Are you OK with "test-coverage"? Then we can merge I think.

cleptric:
@dereuromark Ok, I'll update it later.

dereuromark:
@cleptric PS: In my apps I still prefer to go with the simple one-liner I initially put up as PR: https://github.com/dereuromark/cakephp-sandbox/blob/3.0/composer.json#L86
And in plugins I indeed use the tmp dir: https://github.com/dereuromark/cakephp-shim/blob/master/composer.json#L41

The extra PHP file still seems a bit overhead, but if it is the consensus and needed for this to be merged I am fine with it.
I just dont quite understand the two `--`. Why not just an additional argument?

    composer test-coverage tmp/coverage

etc?

bravo-kernel:
@dereuromark, I think the `--` is a composer requirement for using custom arguments.

> You can also give additional arguments to the script handler by appending -- followed by the handler arguments. e.g. composer run-script post-install-cmd -- --check will pass--check along to the script handler. Those arguments are received as CLI arg by CLI handlers, and can be retrieved as an array via $event->getArguments() by PHP handlers.

src/Console/Coverage.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $io = $event->getIO();
+        $rootDir = dirname(dirname(__DIR__));
+        $coverageDir = $rootDir . '/tmp/';
+
+        if (isset($event->getArguments()[0])) {
+            $coverageDir = $rootDir . '/' . $event->getArguments()[0];
+        }
+
+        exec('phpunit --log-junit ' . $coverageDir . '/coverage/unitreport.xml --coverage-html ' . $coverageDir . '/coverage --coverage-clover ' . $coverageDir . '/coverage/coverage.xml', $out, $status);
```
markstory: You should `escapeshellarg()` these values. People will use paths with spaces and they will be :Cry:

src/Console/Coverage.php
josegonzalez: I think having them here is fine for now. If we wish, the next minor release can move them into a subdirectory.

dereuromark:
Did I mention yet that https://github.com/cakephp/app/pull/486/files would have been so easy, one-liner and super-transparent? :)
This PR here seems to lead nowhere (not sure why tbh), whereas mine really was a dev-only little composer tool to quickly use on the local machine and would never be executed on the server anyway (due to require-dev dependency of phpunit etc).

cleptric:
I just added a simple composer script entry to my skeleton app. So I would not bother to close this, if no one wants to merge it 😛

dereuromark:
And I use mine, if anyone wants they can also just reopen my PR and get it merged :P
So we can probably close.

PR #501, opened 2017-03-04
Disable automatic serialization for JsonView and XmlView

rchavik:
Dumping the content of viewVars by default can pose a security
risk as it may contain sensitive information.

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -60,10 +60,14 @@ class AppController extends Controller
      */
     public function beforeRender(Event $event)
     {
-        if (!array_key_exists('_serialize', $this->viewVars) &&
-            in_array($this->response->type(), ['application/json', 'application/xml'])
-        ) {
-            $this->set('_serialize', true);
-        }
+        /*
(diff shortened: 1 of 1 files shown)


ADmad:
This auto serialization was added as a convenience for easier onboarding for new users. If you set `_serialize` view var in your action this featured is skipped anyway and doesn't pose any risk. I would rather keep the code and add a comment stating that it's just a convenience for quick prototyping and users should remove it and explicitly set `_serialize` once they get started.

inoas:
Can we introduce a `Configure::` setting that keeps the current implementation and can be set to false? I found out how this works by omitting `_serialize`, realizing things worked and then adding it and being explicit.

inoas:
The book says this:
> New in version 3.1.0: You can also set _serialize to true to serialize all view variables instead of explicitly specifying them.

Maybe it should say that that's also the implicit default behavior and can be disabled by using _serialize(['yourExplicitVar])?

rchavik:
Closing as per discussion in dev channel

PR #502, opened 2017-03-20
Removed death-star constraint from composer.json

liviascapin:
As explained in https://thephp.cc/news/2017/02/death-star-version-constraint a '*' constraint is not really a constraint and upgrading to new major versions should be a conscious decision, therefore it should be replaced with a real constraint.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
         "cakephp/cakephp": "3.4.*",
         "mobiledetect/mobiledetectlib": "2.*",
         "cakephp/migrations": "~1.0",
-        "cakephp/plugin-installer": "~0.0.15"
+        "cakephp/plugin-installer": "~1.0"
     },
     "require-dev": {
         "psy/psysh": "@stable",
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
         "cakephp/cakephp": "3.4.*",
         "mobiledetect/mobiledetectlib": "2.*",
         "cakephp/migrations": "~1.0",
-        "cakephp/plugin-installer": "*"
+        "cakephp/plugin-installer": "~0.0.15"
     },
     "require-dev": {
         "psy/psysh": "@stable",


ravage84:
I would generally prefer to use the [caret operator](https://getcomposer.org/doc/articles/versions.md#caret), but anything is better than the dreaded death-star. 😼

ADmad:
Upgrading to major version should surely be a conscious decision made by the developer but in case of plugin installer it's better to rely on the core team's discretion to ensure you don't have any problems with plugin installation :)

If using * is not recommended could we use `@stable` instead?

ravage84:
@ADmad what is your reasoning behind using ``@stable`` instead of a more restrictive version constraint using the Caret or Tilde operator?

If I'm not mistaken, ``@stable`` would mean _any stable version_, probably tending to the newest version possible.
This would include any stable pre-1.0 release and any stable 1.x release, which would be a good thing.
But it would also mean any stable release beyond 1.x, which will probably be not (fully) compatible anymore.

Certainly, such releases are far away for the moment, but I don't see why one should use ``@stable`` when it's clear it _could_ become a problem one day.

markstory:
Would it be helpful if I tag 1.0.0 for the plugin installer? That would let us use `~1.0`

ravage84:
@markstory I don't know whether ``cakephp/plugin-installer``is worthy a ``1.0``, but yes, it would help.

markstory:
Plugin Installer has a 1.0.0 release now.

composer.json
```diff
@@ -9,7 +9,7 @@
         "cakephp/cakephp": "3.4.*",
         "mobiledetect/mobiledetectlib": "2.*",
         "cakephp/migrations": "~1.0",
-        "cakephp/plugin-installer": "*"
+        "cakephp/plugin-installer": "~0.0.15"
```
ravage84: @liviascapin can you change it to ``~1.0``, please? As @markstory [mentioned](https://github.com/cakephp/app/pull/502#issuecomment-287834561).

markstory:
Thanks!

PR #503, opened 2017-04-03
Suggest IdeHelper

dereuromark:
The Bake plugin adds annotation only once, from there on the code gets out of sync with what the IDE needs to fully stay compatible, typehinting, autocomplete, error-detection, ...
The [IdeHelper](https://github.com/dereuromark/cakephp-ide-helper) is therefore just as important after the code is baked and re-baking is not an option anymore. Running it once a week for example can easily merge in the missing annotations or replace outdated ones.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     },
     "suggest": {
         "markstory/asset_compress": "An asset compression plugin which provides file concatenation and a flexible filter system for preprocessing and minification.",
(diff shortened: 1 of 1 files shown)


markstory:
Sounds reasonable to me.

dereuromark:
It is also helpful for PHPStan to stay green (see https://github.com/dereuromark/cakephp-tinyauth/pull/57).

PR #506, opened 2017-05-06
feat: add support for loading .env files for configuration

josegonzalez:
To use, uncomment the following line in your `config/bootstrap.php` file:

    // Configure::load('env', 'default', true);

This will load - in order - the `config/.env` and `config/.env.default` files. If this line is uncommented *but*
there is an environment variable set called `APP_NAME`, then no `.env` files will be loaded/parsed.

A missing `config/.env` file will be ignored when this integration is enabled, though removal of the
`config/.env.default` file will result in errors as it is the last file to be loaded. The first file that is parsed
and loaded will result in those environment variables being set as configuration, and no further
`.env` files will be parsed.

Any customizations *should* be performed in the `config/.env` file, with sane defaults present in the
`config/.env.default`. It is recommended that the corresponding environment variables are loaded through
`config/app.php` and `config/app.php.default` as well, so that production configurations with `APP_NAME`
present will be able to use environment variables as well.

This integration *will not* modify any superglobals, but will instead result in the .env contents
being merged into the configuration provided by `config/app.php`.

Closes cakephp/cakephp#10420

### DIFF
--- a/config/env.php
+++ b/config/env.php
@@ -1,0 +1,55 @@
+<?php
+use Cake\Utility\Hash;
+use josegonzalez\Dotenv\Loader;
+
+$config = [];
+if (!env('APP_NAME')) {
+    $dotenv = new Loader([
+        __DIR__ . DS . '.env',
+        __DIR__ . DS . '.env.default',
+    ]);
+    $dotenv->setFilters([
+        'josegonzalez\Dotenv\Filter\LowercaseKeyFilter',
+        'josegonzalez\Dotenv\Filter\UppercaseFirstKeyFilter',
+        'josegonzalez\Dotenv\Filter\UnderscoreArrayFilter',
+        function ($data) {
+            $keys = [
+                'App.fullbaseurl' => 'App.fullBaseUrl',
+                'Debug' => 'debug',
+                'Email.transport' => 'EmailTransport',
+                'Database.debug.kit' => 'Datasources.debug_kit',
+                'Database.test' => 'Datasources.test',
+                'Database' => 'Datasources.default',
+                'Cache.duration' => null,
+                'Cache.cakemodel' => 'Cache._cake_model_',
+                'Cache.cakecore' => 'Cache._cake_core_',
+            ];
+            foreach ($keys as $key => $newKey) {
+                if ($newKey === null) {
+                    $data = Hash::remove($data, $key);
+                    continue;
+                }
+                $value = Hash::get($data, $key);
+                $data = Hash::remove($data, $key);
+                if ($value !== null) {
+                    $data = Hash::insert($data, $newKey, $value);
+                }
+            }
+
+            foreach ($data['Email'] as $key => $config) {
+                if (isset($config['profile'])) {
+                    parse_str($config['profile'], $output);
+                    $data['Email'][$key] = array_merge($output, $data['Email'][$key]);
+                    unset($data['Email'][$key]['profile'], $output);
+                }
+            }
+
+            return $data;
+        }
+    ]);
+    $dotenv->parse();
+    $dotenv->filter();
+    $config = $dotenv->toArray();
+}
+
+return $config;
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /vendor/*
 /config/app.php
+/config/.env
 /tmp/*
 /logs/*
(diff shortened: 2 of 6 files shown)


config/env.php
```diff
@@ -0,0 +1,55 @@
...
+
+            return $data;
+        }
+    ]);
+    $dotenv->parse();
+    $dotenv->filter();
+    $config = $dotenv->toArray();
+}
+
+return $config;
```
markstory: This is a neat way of integrating dotenv with the existing configuration loaders.

makallio85:
:thumbsup:

josegonzalez:
Needs docs I guess.

ravage84:
Still prefer this as an optional plugin solution. Why must we add a hard dependency that may be 10 % or less of the community use?

I would be totally fine, if we just dedicate a small section in the cookbook for __those__ people who want or need this.

josegonzalez:
@ravage84 it is totally an optional solution. The library for loading `.env` files isn't in the core and is an external, well-tested library. Loading `.env` files is disabled by default, so your application will not change.

For what its worth, I'm one of *those* people.

dereuromark:
I agree with @ravage84 
This should probably be a "suggest" at best, no need to ship this as a require default if only 10% are using it or planning to. A plugin solution at this point seems fair enough.

inoas:
A reason to include would be to be on par with the competition.
http://symfony.com/blog/new-in-symfony-3-3-dotenv-component

Another reason would be if it was best practice - if it is

A Framework, rather than a Library makes choices for you. That helps with teams, working with the framework at other projects/places, getting started, taking the burden to do 100% tooling and env setup from you etc.

I am currently using multiple *_local.php files for different purposes and that's all because I do not want to rely on non-core features for environment setup.

cleptric:
Do we have a source for this 10% claim? I'm more then fine with adding it to this repo, as .env should be something a modern php framework should support out of the box in 2017.

ravage84:
@cleptric no there is not. It's just a number I chose randomly. Replace it with any other low number 😼

ADmad:
While 10% might be a random number I think we can all agree that a very small minority deploys on cloud platform which requires using env vars. I don't think features required by a small minority should be installed by default.

Anyway, my bigger concern is regarding new users. While usage of .env files might be optional, the fact might not be apparent to newbies and it might scare away inexperienced users. Since the composer install script generates `.env` file they might think they must use it. Just the adding of `env()` to current `app.php` has caused confusion for some. I have seen question from some users about what it does and how are they supposed to use it. It's not apparent to them that they could just replace it with static string.

Presenting a modern look for the framework is important but keeping our RAD image and not overwhelming new / inexperienced users with new stuff is equally important. So instead of adding this to the skeleton I would suggest adding a page in manual describing how those who need this feature and setup things up.

ADmad:
@savant We can wire this up for FriendsofCake's skeleton app instead :smile:

ADmad:
> A reason to include would be to be on par with the competition.

I don't see it as a competition between frameworks. It's what suits one's needs. Symfony appeals more to those want an enterprise feel while we present CakePHP as a RAD framework.

>  ... as .env should be something a modern php framework should support out of the box in 2017.

On the other hand we have a user base who still wants to keep using 2.x :)

inoas:
> I don't think features required by a small minority should be installed by default.

I think it should be the default way to configure a few - especially security related - key environment variables such as credentials, salts, etc. It is also a lot easier and less error prone.

> Just the adding of env() to current app.php has caused confusion for some. I have seen question from some users about what it does and how are they supposed to use it. It's not apparent to them that they could just replace it with static string.

No wonder, a half-baked feature in the core config files has to confuse newbies.

> Presenting a modern look for the framework is important but keeping our RAD image and not overwhelming new / inexperienced users with new stuff is equally important. So instead of adding this to the skeleton I would suggest adding a page in manual describing how those who need this feature can setup things up.

That's exactly not RAD.

.env makes it easy to move things that should not go into repos, or are copied around with emptied out values and send by mail to some web host admin (newbies deploying) easy.

RAD and enterprisey should not exclude itself.

ADmad:
An alternate is having composer command or shell which installs the dependency and creates the `.env` thereby avoiding the need to do anything manually (expect run a command) for those who needs this feature.

josegonzalez:
Lord, I thought we got over this hump in the issue... Thanks for chiming in then everyone.

I'll speak with the core devs via core chat to figure out what next steps are here. Please hold off on commenting/reviewing until then. Thanks.

inoas:
Any news?

markstory:
Closing as #526 does a similar thing with fewer layers/transformations.

PR #507, opened 2017-05-09
Use MiddlewareStack as variable.

dereuromark:
Refs https://github.com/cakephp/authentication/pull/126

To be in sync with docs.


ADmad:
It's not a stack, it's a queue. The list is consumed as FIFO, not LIFO.

dereuromark:
We need to adjust the docs then, too.

PR #508, opened 2017-05-10
Update config docs for `.` not being used.

markstory:
Refs cakephp/cakephp#10604


dereuromark:
Should this maybe throw a dev exception here in future versions?

PR #511, opened 2017-05-16
Minor improvements to the PagesController

ravage84:


src/Controller/PagesController.php
```diff
@@ -42,7 +42,9 @@ public function display(...$path)
     {
         $count = count($path);
         if (!$count) {
-            return $this->redirect('/');
```
markstory: But redirect returns a response. In the future we'll actually require `return $this->redirect()`.

ravage84: I see. Then the return type ``void`` is wrong. https://github.com/cakephp/app/pull/511/files#diff-187a1f16b006225e2ca35de472fc4641R36

Then it would be ``null``, as it would fit the return type of  ``Controller::redirect()`` this way.
https://api.cakephp.org/3.4/class-Cake.Controller.Controller.html#_redirect

markstory: Ok. Doc blocks habe a habit of getting wrong when return types are involved.

ravage84: OK, I have adjusted it.

src/Controller/PagesController.php
```diff
@@ -32,8 +32,8 @@ class PagesController extends AppController
     /**
      * Displays a view
      *
-     * @param string ...$path Path segments.
-     * @return void|\Cake\Network\Response
+     * @param string|array ...$path Path segments.
```
dereuromark: How can it still be a string? count() only works on array, does it not?

ravage84: Actually, I wasn't sure myself. Going to fix.

src/Controller/PagesController.php
```diff
@@ -59,9 +59,9 @@ public function display(...$path)
 
         try {
             $this->render(implode('/', $path));
-        } catch (MissingTemplateException $e) {
+        } catch (MissingTemplateException $exception) {
             if (Configure::read('debug')) {
-                throw $e;
+                throw $exception;
             }
             throw new NotFoundException();
```
dereuromark: Should this get the exception as 3rd argument (parent)?

dereuromark: Might require https://github.com/cakephp/cakephp/pull/10649 to land first.

ravage84: I did not want to go too far with my changes. Can be done in another PR.

src/Controller/PagesController.php
ravage84: Fixed.

PR #514, opened 2017-05-25
Changed a few use statements from “Cake\Network\*” to “Cake\Http\*”.

JoshuaLuckers:


config/bootstrap.php
```diff
@@ -60,7 +60,7 @@
 use Cake\Error\ErrorHandler;
 use Cake\Log\Log;
 use Cake\Mailer\Email;
-use Cake\Network\Request;
+use Cake\Http\ServerRequest;
```
ADmad: You also need to change usages like `Request::addDetector()` in the file to `ServerRequest::addDetector()`.

JoshuaLuckers: Thanks for the review. I changed those as well.

PR #517, opened 2017-06-01
Expose DebugKit autoRun option

inoas:
... so that people may disable it.

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -216,7 +216,8 @@ Type::build('timestamp')
 /*
  * Only try to load DebugKit in development mode
  * Debug Kit should not be installed on a production system
+ * You may also set autoRun to false and append ?debug=true to your URLs to toggle DebugKit on.
  */
 if (Configure::read('debug')) {
-    Plugin::load('DebugKit', ['bootstrap' => true]);
+    Plugin::load('DebugKit', ['bootstrap' => true, 'autoRun' => true]);
 }


ravage84:
Refers: https://github.com/cakephp/debug_kit/wiki/Configuration#autorun

I had to search quite a bit to find the documentation on this. I guess the documentation for this could be a bit more visible...

config/bootstrap.php
```diff
@@ -216,7 +216,8 @@
 /*
  * Only try to load DebugKit in development mode
  * Debug Kit should not be installed on a production system
+ * You may also set autoRun to false and append ?debug=true to your URLs to toggle DebugKit on.
  */
 if (Configure::read('debug')) {
-    Plugin::load('DebugKit', ['bootstrap' => true]);
+    Plugin::load('DebugKit', ['bootstrap' => true, 'autoRun' => true]);
```
markstory: I don't think this actually does anything. Instead I think you want autoRun set in Configure.

markstory:
I don't think there is an `autorun` configuration value in 3.x either.

inoas:
You are right, I have no idea why I thought it was working - I did test it - but maybe I overlooked the small icon bottom right.

ravage84:
@markstory I propose to remvoe the Wiki that possibly lead to this confusion in cakephp/debug_kit/issues/532

PR #518, opened 2017-06-14
Improve "error" message on home page

ravage84:
Disabled debug mode locally in my app yesterday and ran into this error message which was not very useful. Luckily, I remembered I disabled the debug mode.

Refers to https://github.com/cakephp/cakephp/issues/10752


lorenzo:
Thanks! This is helpful

PR #520, opened 2017-06-22
Add warning against default serializing setting.

ADmad:
@rchavik has brought this up a while back.

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -61,7 +61,7 @@ class AppController extends Controller
     public function beforeRender(Event $event)
     {
         // Note: These defaults are just to get started quickly with development
-        // and should be used in production. You should instead set "_serialize"
+        // and should not be used in production. You should instead set "_serialize"
         // in each action as required.
         if (!array_key_exists('_serialize', $this->viewVars) &&
             in_array($this->response->type(), ['application/json', 'application/xml'])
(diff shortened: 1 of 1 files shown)


cleptric:
Thx 😀

lorenzo:
Should it maybe be surrounded by `if debug == true else { throw HelpfulException }` ?

davidyell:
I always delete this when I'm creating my project anyway. No idea why it's even in the core. To me it feels like the old Cake magic or a type of 'hack'.

josegonzalez:
@lorenzo That seems extra hacky to me...

lorenzo:
@josegonzalez I just find unhelpful that we provide code that is not meant to be actually there. So, why have it?

inoas:
I am either pro removing or for changing it to warning that goes to the log files in production.
Adding an Exception can result into "deployment-day" and you realize that once you switch from debug true to debug false you suddenly see exceptions you did not expect to see.

davidyell:
There is also a chance that if it's changed to a log, it will not be noticed. If it's logged on every request that will produce a big log file pretty quickly.

ADmad:
> No idea why it's even in the core

It's not in core, it's in the app skeleton which is meant to help people going quickly with development. I had added this in the past as I seen many newbies struggle figuring out what proper vale to set for `_serialize` to get the serialization working.

Hopefully in 4.x we can come up with a better way to handle what to serialize. Settings this view var feels a bit hacky now.

PR #526, opened 2017-07-14
Add dotenv with fewer transformations

markstory:
I've borrowed very liberally from the changes in #506 for this. In talking with @lorenzo he wasn't entirely happy with the number of transformations being done to the environment variables. This approach does no additional transformations and should enable the `.env` file and native environments variables to work exactly the same.

### DIFF
--- a/config/.env.default
+++ b/config/.env.default
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+# Used as a default for seeding
+# config/.env which enables you to use environment variables
+# to configure the aspects of your application that vary by
+# environment. In development .env files are parsed by PHP
+# and set into the environment. This provides a simpler
+# development workflow over standard environment variables.
+export APP_NAME="__APP_NAME__"
+export DEBUG="true"
+export APP_ENCODING="UTF-8"
+export APP_DEFAULT_LOCALE="en_US"
+export SECURITY_SALT="__SALT__"
+
+export CACHE_DURATION="+2 minutes"
+export CACHE_DEFAULT_URL="file://tmp/cache/?prefix=${APP_NAME}_default&duration=${CACHE_DURATION}"
+export CACHE_CAKECORE_URL="file://tmp/cache/persistent?prefix=${APP_NAME}_cake_core&serialize=true&duration=${CACHE_DURATION}"
+export CACHE_CAKEMODEL_URL="file://tmp/cache/models?prefix=${APP_NAME}_cake_model&serialize=true&duration=${CACHE_DURATION}"
+
+export EMAIL_TRANSPORT_DEFAULT_URL=""
+
+export DATABASE_URL="mysql://my_app:secret@localhost/${APP_NAME}?encoding=utf8&timezone=UTC&cacheMetadata=true&quoteIdentifiers=false&persistent=false"
+export DATABASE_TEST_URL="mysql://my_app:secret@localhost/test_${APP_NAME}?encoding=utf8&timezone=UTC&cacheMetadata=true&quoteIdentifiers=false&persistent=false"
+
+export LOG_DEBUG_URL="file://logs?levels[]=notice&levels[]=info&levels[]=debug&file=debug"
+export LOG_ERROR_URL="file://logs?levels[]=warning&levels[]=error&levels[]=critical&levels[]=alert&levels[]=emergency&file=error"
--- a/config/.env.default
+++ b/config/.env.default
@@ -21,5 +21,6 @@ export EMAIL_TRANSPORT_DEFAULT_URL=""
 export DATABASE_URL="mysql://my_app:secret@localhost/${APP_NAME}?encoding=utf8&timezone=UTC&cacheMetadata=true&quoteIdentifiers=false&persistent=false"
 export DATABASE_TEST_URL="mysql://my_app:secret@localhost/test_${APP_NAME}?encoding=utf8&timezone=UTC&cacheMetadata=true&quoteIdentifiers=false&persistent=false"
 
-export LOG_DEBUG_URL="file://logs?levels[]=notice&levels[]=info&levels[]=debug&file=debug"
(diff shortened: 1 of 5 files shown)


markstory:
ping @josegonzalez @lorenzo

lorenzo:
I like this new version, the change also makes it more visible to the developers.

config/bootstrap.php
```diff
@@ -64,6 +64,20 @@
 use Cake\Utility\Inflector;
 use Cake\Utility\Security;
 
+/**
+ * Read .env file if APP_NAME is not set.
+ *
+ * You can remove this block if you do not want to use environment
+ * variables for configuration when deploying.
```
lorenzo: The comment is misleading. The file is. It meant for deploying but is to be used during development.

I understand the intent of your phrase, but it should be made clear what it is actually for.

lorenzo: It would also be nice to have a link to the dotenv docs, in case they want themselves to add transformations

config/bootstrap.php
```diff
@@ -64,6 +64,20 @@
...
+ */
+if (!env('APP_NAME') && file_exists(CONFIG . '.env')) {
+    $dotenv = new \josegonzalez\Dotenv\Loader([CONFIG . '.env']);
```
inoas: This should really be a core repo then IMHO as the skeleton relies on it.
Especially when the skeleton suggests to use .env files by creating them `static::createEnvFile($rootDir, $io);`

Also I am no expert on environment files but would it make sense to prefix all env variables with `CAKEPHP_`

lorenzo: No need to prefixes, it is easy enough to isolate env variables per process.

markstory: How would making it a cakephp owned repository be better? The app skeleton currently depends on mobiledetect which is not a cakephp project as well.

config/bootstrap.php
markstory: What do you think of 

> Environment variables help make deploying your application into different environments easier as each environment can define the configuration unique to that environment.
>
> You can remove this block if you do not use environment variables for configuration.

I'm not sure we should encourage people to use filters, as those wouldn't be applied in production scenarios.

ADmad:
Doesn't the generation of .env by install script and the check for it's existence in bootstrap makes this feature opt out instead of opt in?

config/bootstrap.php
lorenzo: Makes sense to me

config/.env.default
```diff
@@ -0,0 +1,25 @@
...
+export CACHE_CAKECORE_URL="file://tmp/cache/persistent?prefix=${APP_NAME}_cake_core&serialize=true&duration=${CACHE_DURATION}"
+export CACHE_CAKEMODEL_URL="file://tmp/cache/models?prefix=${APP_NAME}_cake_model&serialize=true&duration=${CACHE_DURATION}"
+
+export EMAIL_TRANSPORT_DEFAULT_URL=""
+
+export DATABASE_URL="mysql://my_app:secret@localhost/${APP_NAME}?encoding=utf8&timezone=UTC&cacheMetadata=true&quoteIdentifiers=false&persistent=false"
+export DATABASE_TEST_URL="mysql://my_app:secret@localhost/test_${APP_NAME}?encoding=utf8&timezone=UTC&cacheMetadata=true&quoteIdentifiers=false&persistent=false"
+
+export LOG_DEBUG_URL="file://logs?levels[]=notice&levels[]=info&levels[]=debug&file=debug"
+export LOG_ERROR_URL="file://logs?levels[]=warning&levels[]=error&levels[]=critical&levels[]=alert&levels[]=emergency&file=error"
```
lorenzo: Should all the lines here be commented out?

I can see this being a source of confusion as things defined here will override stuff in app.php. Someone will get surprised when they see their settings not being applied

markstory:
@ADmad I could remove the generation of the `config/.env` file or comment out the .env file loading?

config/.env.default
markstory: Ok. I'll comment them out.

josegonzalez:
@markstory dont create the .env file and add a comment to the top of .env.default to say that this isn't loaded without the `bootstrap.php` change?

markstory:
Updated the Installer and .env.default comments.

markstory:
@lorenzo @josegonzalez @ADmad How is this looking now?

config/bootstrap.php
```diff
@@ -64,6 +64,20 @@
...
+    $dotenv->parse()
+        ->putenv()
+        ->toEnv()
+        ->toServer();
```
josegonzalez: Not a fan of forcing them to be in the env, but I guess if we don't want crazy transformations, thats the best we can do.

src/Console/Installer.php
```diff
@@ -186,10 +200,40 @@ public static function setSecuritySalt($dir, $io)
...
+    /**
+     * Set the APP_NAME value in a given file
+     *
+     * @param string $dir The application's root directory.
+     * @param \Composer\IO\IOInterface $io IO interface to write to console.
+     * @param string $appName app name to set in the file
+     * @param string $file A path to a file relative to the application's root
+     * @return void
+     */
+    public static function setAppNameInFile($dir, $io, $appName, $file)
```
ADmad: Where is this method called from?

robertpustulka:
Great, maybe this will convince me to start using env finally :smiley:

saeideng:
I hope to use this in next time
Looks good at this time :smile:

config/bootstrap.php
inoas: Agreed however many apps could probably do without mobiledetect whereas for env we could see if it becomes the new default including docs in the book. 

Well but once that happens it can still be moved it inside the cakephp repo.

config/bootstrap.php
markstory: Forcing into env also allows dotenv to work like real env vars.

josegonzalez: Thought they get leaked into error reporting if you're not careful ;)

PR #529, opened 2017-07-31
fetch content needs 'echo'

littleylv:
refs #513 

change `$this->fetch('content');` to `echo $this->fetch('content');`


dereuromark:
Thank you!

Issue #510, opened 2017-05-16
Inconvenient .gitignore for contributing


ravage84:
The [.gitignore](https://github.com/cakephp/app/blob/3.4.2/.gitignore) file in its current state is totally inconvenient for contributing.

Due to a few changes ([1](https://github.com/cakephp/app/commit/cd7efc5c1679f3f4d0c927a5386bd0d71c98ae1e), [2](https://github.com/cakephp/app/commit/9509d503eacc7d7e33d36d7b1c1ae02388ea7c9e)), the .gitignore became from [totally convenient for contributing](https://github.com/cakephp/app/commit/5d6a60c84c5958c2061c01ea89fdf25fc19ab388) to totally inconventient as now one gets several files prompted for being versioned that do not need to be versioned for an open source project.

Those files are:
- .idea/* (PhpStorm project folder)
- bin/boris & bin/boris.bat
- composer.lock
- plugins/DebugKit/*

I understand we have to choose between being optimized for open source contributions or for project usage. I would prefer the first, as the inconvenience is bigger right now. Currently every open source contributor has to **make sure he/she does not commit one of the files above every time** he/she wants to contribute. If we would optimise for OSS contributions, the only thing a developer has to do after setting up his/her new app **is to adjust the .gitignore to his/her very specific needs once**.

Also, I know some say IDE specific folders such as .idea and typical PHP tool files such as *.phar should be ignored through one's global .gitignore file.  This is all fine and dandy, but the truth I observed over the last few years is that most developers do not set their global .gitignore file properly (including me). It's much more convenient if this is setup by project/repo.

For example, internally we started to version some files in the .idea folder for internal projects recently, while we do not version them for OSS projects.

dereuromark:
Actually, excluding 

    composer.lock

by default is a bad practice, also for app skeleton (as boilerplate for a new project).

.idea is pretty proprietary (but I would be fine adding this again, doesnt hurt).
I dont know about the other two.

ravage84:
@dereuromark  I do not see why excluding the ``composer.lock`` **in this particular case** is bad practise. I can agree it is in other cases, though.

But here, we have two clear usage scenarios:

a) OSS contributions aka the "CakePHP App Template repo"
b) project usage aka internal

For a) we can either add the ``composer.lock`` to be versioned and then we do not need to exclude it or we do not add it but should then exclude it as **it will be prompted to be added every time**, otherwise.

[...]

PR #532, opened 2017-08-18
.gitignore: re-add useful defaults

inoas:
re-add useful defaults
Ref: https://github.com/cakephp/app/issues/510#issuecomment-311345620

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,43 @@
-/vendor/*
+# CakePHP specific files #
+##########################
 /config/app.php
-/tmp/*
+/config/.env # josegonzalez/php-dotenv
 /logs/*
+/tmp/*
+/tmp/bake/*
+!/tmp/bake/empty
+/tmp/cache/models/*
+!/tmp/cache/models/empty
+/tmp/cache/persistent/*
+!/tmp/cache/persistent/empty
+/tmp/cache/views/*
+!/tmp/cache/views/empty
+/tmp/sessions/*
+!/tmp/sessions/empty
+/tmp/tests/*
+!/tmp/tests/empty
+/vendor/*
+
+# OS generated files #
+######################
+.DS_Store
+.DS_Store?
+._*
+.Spotlight-V100
+.Trashes
+Icon?
+ehthumbs.db
+Thumbs.db
+
+# Tool specific files #
+#######################
+*~ # vim
+*.swp # vim
+*.swo # vim
+*.sublime-* # sublime text
+*.tmlanguage.cache # sublime text & textmate
+*.tmPreferences.cache # sublime text & textmate
+*.stTheme.cache # cache files for sublime text
+.idea/* # JetBrains, aka PHPStorm, IntelliJ IDEA
+/nbproject # NetBeans
+.sass-cache/ # Sass preprocessor
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,6 @@ Thumbs.db
 *.tmPreferences.cache # sublime text & textmate
 *.stTheme.cache # cache files for sublime text
 .idea/* # JetBrains, aka PHPStorm, IntelliJ IDEA
-nbproject # NetBeans
+nbproject/* # NetBeans
 .vscode # Visual Studio Code
 .sass-cache/ # Sass preprocessor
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,4 +1,43 @@
-/vendor/*
+# CakePHP specific files #
+##########################
 /config/app.php
-/tmp/*
+/config/.env # josegonzalez/php-dotenv
 /logs/*
+/tmp/*
+/tmp/bake/*
```
dereuromark: if tmp is already in, please do not add also all subfolders as well.
Lets keep it simple when adding this again please.

inoas:
These are debate-able: https://github.com/cakephp/app/pull/532/files#diff-a084b794bc0759e7a6b77810e01874f2R32

However I think we should have sufficient out of the box support for:
- vim
- textmate
- sublime
- atom
- PHPStorm / JetBrains
- Eclipse/Aptana
- Xcode
- Netbeans
- Emacs
- Common cache directories for transpilers (Elm, Coffescript, Sass, Scss, Less, Stylus)

If we cannot find a way to do that I'd say we should either support nothing or just the bare minimum: vim (Disclaimer: I am a textmate user).

bravo-kernel:
Visual Studio Code perhaps?

.gitignore
inoas: I think the tools are really the thing we should discuss.
I don't think adding `/tmp/*` is good practise. The `empty` files should stay and without those extra directives, they would not.

p.s.: Personally I am using `.keep` however as that is much better than `empty` (says more about why the file exists, is hidden by default, etc)

inoas:
> Visual Studio Code perhaps?

Name the files :)

bravo-kernel:
`.vscode` folder

.gitignore
dereuromark: The deploy scripts live take care if it (or should!), and locally they are auto.-created.
No need IMO to do anything else here.

.gitignore
```diff
@@ -1,4 +1,43 @@
...
+#######################
+*~ # vim
+*.swp # vim
+*.swo # vim
+*.sublime-* # sublime text
+*.tmlanguage.cache # sublime text & textmate
+*.tmPreferences.cache # sublime text & textmate
+*.stTheme.cache # cache files for sublime text
+.idea/* # JetBrains, aka PHPStorm, IntelliJ IDEA
+/nbproject # NetBeans
```
ravage84: The leading slash could and probably should be removed as it would not exclude such a NetBeans folder in a sub directory, right now.

.gitignore
inoas: I think it is much better if they exist beforehand to avoid any sources of permission problems creating directories. Also they have no side-effects as they are mandatory by the CakePHP framework. When creating this PR I assumed there would be a backslash on adding so much editor support, rather than this, very basic sane default.

I'll fix some things and if there are no other voices I will remove the tmp directory ignores.

ADmad:
`.settings` for Eclipse.

.gitignore
```diff
@@ -26,6 +26,7 @@ Thumbs.db
 *.tmlanguage.cache # sublime text & textmate
 *.tmPreferences.cache # sublime text & textmate
 *.stTheme.cache # cache files for sublime text
+.setting # Eclipse
```
ADmad: `.settings` with an "s"

ceeram:
personally i dont understand why a project should add non-project files and folders to its `.gitignore` 
I my OS or IDE generates files i want to be ignored in git projects, i would add those to my global git ignore instead of having those repeated in every git project i work on.

ADmad:
@ceeram I have brought up the same argument in the past but gave up since no one else seemed to share my POV :slightly_smiling_face:

dereuromark:
@ceeram True, but for many it can be the first of a project to work on, and this makes their life a bit easier, it also doesnt hurt if the gitignore file contains them again (we all can just skip those lines with our eyes).

ceeram:
re-add? was this ever even in the .gitignore before?

dereuromark:
Yes, someone removed them all a while ago.

markstory:
@ceeram I'm with @dereuromark on this. New developers will likely not know about global gitignores. Experienced folks might not know about them too.

ravage84:
@ceeram from my related [issue](https://github.com/cakephp/app/issues/510#issue-229095809):

> Also, I know some say IDE specific folders such as .idea and typical PHP tool files such as *.phar should be ignored through one's global .gitignore file. This is all fine and dandy, but the truth I observed over the last few years is that most developers do not set their global .gitignore file properly (including me). It's much more convenient if this is setup by project/repo.

inoas:
So is there need to add .phar or are we fine now?

PR #533, opened 2017-08-18
unify early platform requirement checks

inoas:
Unify early platform requirement checks in front of both "front contollers", index.php and cake.php

See: https://github.com/cakephp/app/pull/438#issuecomment-323290506

### DIFF
--- a/bin/cake.php
+++ b/bin/cake.php
@@ -16,17 +16,8 @@
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
 
-$minVersion = '5.6.0';
-if (file_exists('composer.json')) {
-    $composer = json_decode(file_get_contents('composer.json'));
-    if (isset($composer->require->php)) {
-        $minVersion = preg_replace('/([^0-9\.])/', '', $composer->require->php);
-    }
-}
-if (version_compare(phpversion(), $minVersion, '<')) {
-    fwrite(STDERR, sprintf("Minimum PHP version: %s. You are using: %s.\n", $minVersion, phpversion()));
-    exit(-1);
-}
+// Check platform requirements
+require dirname(__DIR__) . '/config/requirements.php';
 
 require dirname(__DIR__) . '/vendor/autoload.php';
 include dirname(__DIR__) . '/config/bootstrap.php';
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -13,31 +13,6 @@
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
 
-// You can remove this if you are confident that your PHP version is sufficient.
-if (version_compare(PHP_VERSION, '5.6.0') < 0) {
-    trigger_error('Your PHP version must be equal or higher than 5.6.0 to use CakePHP.', E_USER_ERROR);
-}
-
-/*
- *  You can remove this if you are confident you have intl installed.
- */
-if (!extension_loaded('intl')) {
-    trigger_error('You must enable the intl extension to use CakePHP.', E_USER_ERROR);
-}
-
-/*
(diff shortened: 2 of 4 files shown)


bin/cake.php
```diff
@@ -16,17 +16,8 @@
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
 
-$minVersion = '5.6.0';
-if (file_exists('composer.json')) {
```
inoas: While this was a nice way of checking, it is also a slow way (file_exists check, file_get_contents, json_decode, preg_replace)

bin/cake.php
```diff
@@ -16,17 +16,8 @@
...
-    $composer = json_decode(file_get_contents('composer.json'));
-    if (isset($composer->require->php)) {
-        $minVersion = preg_replace('/([^0-9\.])/', '', $composer->require->php);
-    }
-}
-if (version_compare(phpversion(), $minVersion, '<')) {
-    fwrite(STDERR, sprintf("Minimum PHP version: %s. You are using: %s.\n", $minVersion, phpversion()));
```
inoas: I can reintroduce fwrite to STDERR if you think this is better than just trigger_error() in requirements.php

bin/cake.php
dereuromark: True, in cli this is irrelevant, for web usage it might have an impact..

bin/cake.php
dereuromark: trigger_error is fine.

bin/cake.php
inoas: I had this thought too, and i think even in CLI this is relevant if you write a cli based app with tons of calls. So it is aside the elegance of reading from composer ... better anyway.

config/requirements.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+if (!extension_loaded('intl')) {
+    trigger_error('You must enable the intl extension to use CakePHP.' . PHP_EOL, E_USER_ERROR);
+}
+
+/*
+ * You can remove this if you are confident you have mbstring installed.
+ */
+if (!extension_loaded('mbstring')) {
+    trigger_error('You must enable the mbstring extension to use CakePHP.' . PHP_EOL, E_USER_ERROR);
```
inoas: I have added  . PHP_EOL so that the CLI (and windows CLI) do well.

config/bootstrap.php
```diff
@@ -13,31 +13,6 @@
...
- */
-if (!extension_loaded('mbstring')) {
-    trigger_error('You must enable the mbstring extension to use CakePHP.', E_USER_ERROR);
-}
-
-/*
- * Configure paths required to find CakePHP + general filepath
- * constants
- */
-require __DIR__ . '/paths.php';
```
markstory: This looks like it might be important.

inoas: Sorry!

gonzela2006:
@inoas I think we also should check if `openssl` extension is loaded.

inoas:
@gonzela is that a requirement and how/when is it being used?

How do you check for it locally?

markstory:
@gonzela2006 openssl is not a hard requirement for basic functionality though.

inoas:
And then not sure if there are multiple ways... https://en.m.wikipedia.org/wiki/Comparison_of_TLS_implementations

inoas:
Can composer do php extension suggestions?

markstory:
@inoas Yes. ext-openssl is already there.

inoas:
I can't judge if there are, and in case, which drop-in-replacement/version is the best (open, vs libre, etc) but that's for another discussion, and we are done here, then @ non-hard-dependency. Good for you @gonzela2006 ?

PR #534, opened 2017-08-18
add conventional log scopes

inoas:
Added scopes for queries and 404 as these generally pollute the errors and debug logs the most and the benefit to have them separate to look at is high...

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -286,13 +286,6 @@ return [
      * Configures logging options
      */
     'Log' => [
-        '404' => [
-            'className' => 'Cake\Log\Engine\FileLog',
-            'path' => LOGS,
-            'file' => '404',
-            'url' => env('LOG_NOT_FOUND_URL', null),
-            'scopes' => ['404'],
-        ],
         'debug' => [
             'className' => 'Cake\Log\Engine\FileLog',
             'path' => LOGS,
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -236,7 +236,7 @@ return [
             'timezone' => 'UTC',
             'flags' => [],
             'cacheMetadata' => true,
-            'log' => true,
+            'log' => false,
 
             /**
              * Set identifier quoting to true if you are using reserved words or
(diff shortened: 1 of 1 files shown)


config/app.default.php
```diff
@@ -286,19 +286,35 @@
      * Configures logging options
      */
     'Log' => [
+        '404' => [
+            'className' => 'Cake\Log\Engine\FileLog',
+            'path' => LOGS,
+            'file' => '404',
+            'url' => env('LOG_NOT_FOUND_URL', null),
+            'scopes' => ['404'],
```
markstory: This isn't a scope/log we write to by default.

inoas: But could we? We can rename it to NotFound? Otherwise I can remove it. However it helped me quite a lot spotting missing files VS not polluting my other log files.

inoas:
Removed the 404 scope

markstory:
If people want to turn off 404 logging by default, we could update the `Error.skipLog` config option

inoas:
skipLog I think that's good for production (maybe?), but not for development where I would want to always see 404.

inoas:
GTM?

config/app.default.php
```diff
@@ -236,7 +236,7 @@
             'timezone' => 'UTC',
             'flags' => [],
             'cacheMetadata' => true,
-            'log' => false,
+            'log' => true,
```
markstory: This is not a good default to have. People will leave this on by accident and soak up all their disk.

inoas: Well the intention was to help with getting people going with development and discovering the query log.

In a similar way wouldn't you say we need to disable 4xx logging, then, too?

If reading the debug flag was available here... I'd say set it to true on debug, false on debug false?

markstory: Wouldn't people be using debugkit out of the box which includes/enables query logging?

inoas: While I like and debug kit for many things, I prefer to see the sql logging response right away on a secondary screen when working with the query builder; also this is useful when working with Postman.app where we don't have DebugKit - in the very long run (4.0, 5.0) it would be nice if DebugKit had a remote connection option.

Anyway I will remove it from this PR.

inoas:
Anything left to be done?

Issue #535, opened 2017-08-19
Improve app/config/schema/sessions.sql


inoas:
Suggestion:

```sql
CREATE TABLE `sessions` (
  `id` char(40) CHARACTER SET ascii COLLATE ascii_bin NOT NULL,
  `created` datetime DEFAULT CURRENT_TIMESTAMP,
  `modified` datetime DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  `data` blob DEFAULT NULL,
  `expires` int(11) unsigned DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

Ref: https://book.cakephp.org/3.0/en/development/sessions.html#database-sessions
Ref: https://github.com/cakephp/app/blob/master/config/schema/sessions.sql

**Open Questions**

- Did the sessions.sql schema ran on PostGreSQL before?
- Should it? Should/could we supply multiple sessions.sql-files depending on the adapter?
- What is the max length the ID field can take... is it really char(40)?

inoas:
Also in DatabaseSession.php:

```
    /**
     * Helper function called on write for database sessions.
     *
     * @param int $id ID that uniquely identifies session in database
     * @param mixed $data The value of the data to be saved.
     * @return bool True for successful write, false otherwise.
     */
    public function write($id, $data)
    {
```

... but `$id` is really a string, isn't it?

markstory:
> Did the sessions.sql schema ran on PostGreSQL before?

I don't think so, as it has been a blob field for quite some time.

> ... but $id is not of type int, but of string, isn't it?

Yes it is.

inoas:
`blob`? it is still text in cakephp/app just not in cakephp/docs...

PRs are open, ... closing.

PR #537, opened 2017-08-20
Improve sessions.sql

inoas:
* more strict types
  - table => utf-8 charset
  - primary-key => ascii_bin
  - data => blob
  - expires => unsigned & could in theory be set to NULL (does not expire)
* added created and modified which are set automatically without having to create the table object (but you can and the DatabaseSession handler will take a `SessionsTable.php` if it can find it). These fields help with server side invalidation.

Ref: https://github.com/cakephp/app/issues/535

### DIFF
--- a/config/schema/sessions.sql
+++ b/config/schema/sessions.sql
@@ -5,9 +5,11 @@
 # Redistributions of files must retain the above copyright notice.
 # MIT License (https://opensource.org/licenses/mit-license.php)
 
-CREATE TABLE sessions (
-  id char(40) NOT NULL,
-  data text,
-  expires INT(11) NOT NULL,
-  PRIMARY KEY  (id)
-);
+CREATE TABLE `sessions` (
+  `id` char(40) CHARACTER SET ascii COLLATE ascii_bin NOT NULL,
+  `created` datetime DEFAULT CURRENT_TIMESTAMP,
+  `modified` datetime DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  `data` blob DEFAULT NULL,
+  `expires` int(11) unsigned DEFAULT NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;


ravage84:
Was this SQL file MySQL specific before? Now it is due to the use of MySQL specific features, such as InnoDB...

chinpei215:
Since there was `int(11)`, it wouldn't support other databases originally.

ravage84:
@chinpei215 Thanks.

Isn't this file what is shown [here](https://book.cakephp.org/3.0/en/development/sessions.html#database-sessions) too?

chinpei215:
Since it says,
> This configuration will require a database table to be added with **at least these fields**:

we would have to change the above sentence if we want to change the scheme declaration in the page.

lorenzo:
We should maybe have a migration instead that can be copied into your own migrations folder

config/schema/sessions.sql
```diff
@@ -5,9 +5,11 @@
...
-  data text,
-  expires INT(11) NOT NULL,
-  PRIMARY KEY  (id)
-);
+CREATE TABLE `sessions` (
+  `id` char(40) CHARACTER SET ascii COLLATE ascii_bin NOT NULL,
+  `created` datetime DEFAULT CURRENT_TIMESTAMP,
+  `modified` datetime DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  `data` blob DEFAULT NULL,
+  `expires` int(11) unsigned DEFAULT NULL,
```
inoas: The digits limitation of the integer types in SQL are just for zero-filling, if that exists. However I will still change it, as it won't hurt and the current max unix timestamp is 32bit int (https://en.wikipedia.org/wiki/Unix_time)

ravage84:
I agree with @lorenzo . Wanted to suggest that, anyway.

inoas:
@lorenzo I don't think we should remove the easy way to create the sessions table. However I would welcome a migration-way to do it.

lorenzo:
No, I'm not saying to remove this, just add another way.

PR #542, opened 2017-08-25
Keep env vars commented.

ADmad:
Making users uncomment them ensures they will hopefully notice the default values set and update as required.
During a live coding session an individual copied `.env.default` to `.env` without noticing that `/tmp/cache` was now being used as cache folder and kept trying to modifying perms of `tmp/cache` inside the app folder.


markstory:
Makes sense to me. This makes each env var a user wants to use a more clear opt-in step.

ADmad:
Closing in favor of #544

PR #544, opened 2017-08-25
Making usage of `.env` file opt-in instead of opt-out.

ADmad:
Also comment out important vars in `.env.default` so that after copying it to `.env`
users have to explicitly uncomment and set proper values.

### DIFF
--- a/config/.env.default
+++ b/config/.env.default
@@ -4,7 +4,8 @@
 # the aspects of your application that vary by
 # environment.
 #
-# To use this file, first copy it into `config/.env`.
+# To use this file, first copy it into `config/.env`. Also ensure the related
+# code block for loading this file is uncommented in `config/boostrap.php`
 #
 # In development .env files are parsed by PHP
 # and set into the environment. This provides a simpler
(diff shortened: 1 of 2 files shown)


inoas:
So the env loader AND most of the env things are outcommented? Is that really required (goes against convention out of the box concept?)

markstory:
I agree with @inoas do both the loader and variables both need to be commented? Doing just the loader should be enough to create an opt in experience.

ADmad:
Yes commenting out the env vars is also necessary.

During the live coding session Elena copied .env.default to .env without noticing that the default path was `/tmp/cache` and started having permissions issues. She then kept trying to modifying perms of tmp/cache inside the app folder. She could never figure out her mistake as had to start over again. It's very likely that another newbie might make the same mistake.

Having the values commented makes it more likely that users will notice and change the defaults as required when uncommenting.

josegonzalez:
We could also ensure that the tmp dir ends up being what it would be by default?

markstory:
Perhaps `/tmp/cache` is a bad default to have like @josegonzalez said. Is there a way to make the env variables dependent on constants or relative paths?

ADmad:
Today a user on IRC had this same issue. Not realising that the env file defaults the temp folder to root `/tmp` instead of app's `tmp`.

I suggest merging this and later someone can figure out how to make it default to app's tmp if needed.

PR #547, opened 2017-09-02
Update composer.json

ADmad:
- Use recommended caret operator instead of tilde.
- Add config to keep package list sorted for viewing pleasure.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,16 +7,16 @@
     "require": {
         "php": ">=5.6",
         "cakephp/cakephp": "3.5.*",
-        "mobiledetect/mobiledetectlib": "2.*",
-        "cakephp/migrations": "~1.0",
-        "cakephp/plugin-installer": "~1.0",
-        "josegonzalez/dotenv": "2.*"
+        "cakephp/migrations": "^1.0",
+        "cakephp/plugin-installer": "^1.0",
+        "josegonzalez/dotenv": "2.*",
+        "mobiledetect/mobiledetectlib": "2.*"
     },
     "require-dev": {
-        "psy/psysh": "@stable",
-        "cakephp/debug_kit": "~3.2",
-        "cakephp/bake": "~1.1",
-        "cakephp/cakephp-codesniffer": "^3.0"
+        "cakephp/bake": "^1.1",
+        "cakephp/cakephp-codesniffer": "^3.0",
+        "cakephp/debug_kit": "^3.2",
+        "psy/psysh": "@stable"
     },
     "suggest": {
         "markstory/asset_compress": "An asset compression plugin which provides file concatenation and a flexible filter system for preprocessing and minification.",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -7,16 +7,16 @@
     "require": {
         "php": ">=5.6",
         "cakephp/cakephp": "3.5.*",
```
inoas: Why is this not ^3.5.0?

ADmad: Because that would automatically upgrade to 3.6.0 when it's released and we don't want that. Even though we maintain BC between minor release based on past experience it's best we let devs upgrade to next minor themselves are referring to the migration guide.

inoas: So we are out of sync with semver? For the dependencies we assume things are BC, too.
It requires the developer to run composer update anyway because composer lock will install the specific version anyway via composer install. I think having less platform spread by ^3.6.0 default will help the community and the progress of cakephp vs the old way where people are still now relying on cakephp2 etc.

inoas: > Not everyone is familiar enough with how exactly composer conversioning works. While production installs might not be affected having unexpected change during development is also not acceptable.

Those tinkering with CakePHP internals, programming against specific Interfaces, extending and building their own core classes surely know how `composer update`<sup>1</sup> vs `install` works, don't you think?

Then those who do not tinker may know or not know about the difference, it doesn't matter: They are not affected, are they?

On the other hand the community is affected negatively by not using caret here, IMHO:
- Upgrading is harder as it has to happen on purpose later and not just by accident for newbies. Upgrading later means you are more likely to run into multiple problems at the same time, making upgrading a hassle, as the sum of potential problems goes up.
- Minimizing platform spread *by default* is a good thing, isn't it? Commonly community support has to target multiple minor versions instead of only multiple (latest stable) major versions. You have to ask specifically for the CakePHP 3.x version all time because you have to assume people are still using... say 3.1.

Are you still convinced that caret is inferior here especially considering those who tinker against the internals know `composer update`?

***
<sup>1</sup> It should really be named `composer upgrade` (like in apt-get upgrade) :/.

hmic: 👍

markstory: @inoas When we were using `^` we'd get a few bug reports on each minor branch release related to being upgraded to a version they didn't want. #419 is when we moved to the `3.x.*` style requirement. I can't remember if 3.3 was a bumpy release, but that may have contributed to the version constraint switch.

As for minimizing spread, and encouraging more frequent upgrades. These are things that we have no control over, as people would need to run `composer update` in all scenarios which we can't force them to do. For those interested in upgrading the release notes include the compose command to switch release branches.

The more conservative version constraint is intended to reduce surprises for new/less experienced developers. Like you said 'tinkerers' are very likely to know how to operate composer and can easily change their constraints.

inoas: Point taken @ tinkerers can change it anyway.

Issue #549, opened 2017-09-03
Move paths to a constant or class variable


angelxmoreno:
This is a (multiple allowed):

* [ ] bug
* [X] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Application Skeleton Version: 3.5.0
* Platform and Target: composer create-project

### What you did
I extended `App\Console\Installer` in order to create a custom installer with additional folders and permissions. 

### What happened
It was not fun and my code was not DRY.

### What you expected to happen
I expected paths to be a Class variable that I could simply override. Instead, the paths are part of the method body; thus forcing me to repeat the entire method when all i simply needed was add to the existing array of paths.
P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

angelxmoreno:
If this makes sense to the core devs, I'd like to introduce something like:
```php
namespace App\Console;

class Installer
{
    const WRITABLE_DIRS = [
            'logs',
            'tmp',
            'tmp/cache',
            'tmp/cache/models',
            'tmp/cache/persistent',
            'tmp/cache/views',
            'tmp/sessions',
            'tmp/tests'
        ];

    public static function createWritableDirectories($dir, $io)
    {
        foreach (static::WRITABLE_DIRS as $path) {
            $path = $dir . '/' . $path;
            if (!file_exists($path)) {
                mkdir($path);
                $io->write('Created `' . $path . '` directory');
            }
        }
    }
}

markstory:
Using a constant works for me.

angelxmoreno:
I'll make a pr

markstory:
Closing as a pull request is open now.

PR #550, opened 2017-09-04
Moved Installer writeable paths to constant

angelxmoreno:
- Installer paths are now a constant that can more easily be overridden by extending `App\Console\Installer`


angelxmoreno:
Solves #549

markstory:
Thanks :+1:

Issue #489, opened 2017-01-23
Upgrade foundation


dereuromark:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

### What you did
Since there is no version notice anywhere about what the default template ships with in regard of Foundation, the Foundation docs lead to believe one can use http://foundation.zurb.com/sites/docs/callout.html etc.
But in fact we are using such an old version that none of those docs really apply anymore.
The header breaks, and also things like `em` highlighting as mentioned in Foundation docs doesn't work.

### What you expected to happen
We should upgrade to Foundation 6.x (6.3 ideally) or more clear point out the exact version we ship with, so people don't trip over this when trying ou the default app template.

lorenzo:
sure, do you think you have time to change the stylesheet?

dereuromark:
I don't, maybe someone can do that, including a small release version note in the base.css so it is also more easy to figure out the version which was "hardcoded" into the app webroot.

ADmad:
Besides the old version the more likely cause of the issue is we use a custom build of only a subset of the foundation classes not the full framework.

dereuromark:
Can't we use just the fullstack min file for simplicity reasons? The file size shouldn't matter here.

ADmad:
Instead of investing time in this I would rather we update the documentation stating that unless a user i gonna just do prototyping using the default bake templates, they should replace the included css files with their own.

dereuromark:
:-1: ADmad
We should ship with a proper default. And we should spend time here instead of saying "wont fix".
Users should be able to play with the current "default template" shipped - and we sure don't want to start our own Foundation docs just because of this, or say "sry, we have a broken layouting framework here".
That is not the reasonable thing IMO.

cleptric:
Also, we should know what @phpnut has in mind regarding the new layout. Will the new one still be based on Foundation? Maybe a switch to bootstrap?
If we can agree on something I'm, more than fine with doing a PR to upgrade Foundation.

silakanveli:
I have included whole minified foundation 5.5.3, seems to be working. I had problem getting some foundation components working originally. Would be fantastic if you could include foundation 6 by default.

cleptric:
I talked to @phpnut and the new design will still use Foundation.
So I can start to implement Foundation 6.

ADmad:
[...]

PR #552, opened 2017-09-15
Specify version for clarification

dereuromark:
People ask this otherwise.

Refs https://github.com/cakephp/app/issues/489 as possible v6+ upgrade


markstory:
Good idea!

PR #554, opened 2017-10-14
Use UTF-8 Full Unicode / 4 Byte by default

inoas:
- [x] Enhancement

This change affects MySQL and MariaDB.

Otherwise utf8 support will mean your data gets truncated once a 4-byte utf8 symbol gets inserted such as `'😃'`

- Issue with more Info: https://github.com/cakephp/cakephp/issues/11324#issuecomment-336672560

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -289,7 +289,7 @@ return [
             'cacheMetadata' => true,
             'quoteIdentifiers' => false,
             'log' => false,
-            //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
+            // 'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
             'url' => env('DATABASE_TEST_URL', null),
         ],
     ],
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -232,7 +232,8 @@ return [
             'username' => 'my_app',
             'password' => 'secret',
             'database' => 'my_app',
-            'encoding' => 'utf8',
+            // Non MariaDB/MySQL equivalent is 'utf8'
+            'encoding' => 'utf8mb4',
             'timezone' => 'UTC',
             'flags' => [],
             'cacheMetadata' => true,
@@ -272,7 +273,8 @@ return [
             'username' => 'my_app',
             'password' => 'secret',
             'database' => 'test_myapp',
-            'encoding' => 'utf8',
+            // Non MariaDB/MySQL equivalent is 'utf8'
+            'encoding' => 'utf8mb4',
             'timezone' => 'UTC',
             'cacheMetadata' => true,
             'quoteIdentifiers' => false,
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -241,10 +241,6 @@ return [
             'database' => 'my_app',
             /*
              * You do not need to set this flag to use full utf-8 encoding (internal default since CakePHP 3.6).
-             * If you do these are the recommendations:
-             * 'encoding' => 'utf8mb4' // On MariaDB/MySQL for full UTF-8 encoding (4-byte)
-             * 'encoding' => 'utf8mb3' // On MariaDB/MySQL for limited UTF-8 encoding (3-byte, not recommended)
-             * 'encoding' => 'utf8' // On any other RDBMS for full 4-byte UTF-8 encoding (4-byte)
              */
             //'encoding' => 'utf8mb4',
             'timezone' => 'UTC',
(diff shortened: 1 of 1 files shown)


inoas:
Another way would be to remove the encoding flag entirely from the default app.php - or at least outcomment it.

Note: I assume, but did not test (don't have postgres box around atm), that if you use postgres with utf8mb4 it will simply fail. Then you change your app.php config to utf8 and you are good to go again.

utf8 should be the default anyway and the drivers can/will point to the right encoding (utf8) already (Fix for MySQL/MariaDB is here: https://github.com/cakephp/cakephp/pull/11328)

markstory:
I am not sure it is a good idea to use a default encoding that will break on 3/4 db drivers we have. Changing the default just for Mysql and commenting out the encoding key would let people know that the option exists without overriding the defaults.

inoas:
* OT: We could use opt-in analytics much like brew does on the cake home.ctp? Having a flag like --with-usage-statistics on composer create?

inoas:
(Hit the button accidentially)

dereuromark:
Raising awareness is good, changing defaults maybe makes non MySQL people trip over.

inoas:
As long as the default-out-of-the box does not accidentally and suddenly during runtime break queries 👍 ... changing.

inoas:
Note that the minimum MySQL Version to support real utf-8 (including emojis, in URLs and stuff like that) is 5.5.3+. See https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb4.html.
Before that there was just `utf8` which afterwards maps to `utf8mb3`.

config/app.default.php
```diff
@@ -232,7 +237,14 @@
...
             'database' => 'my_app',
-            'encoding' => 'utf8',
+            /**
+             * You do not need to set this flag to use full utf-8 encoding (internal default since CakePHP 3.6).
+             * If you do these are the recommendations:
+             * 'encoding' => 'utf8mb4' // On MariaDB/MySQL for full UTF-8 encoding (4-byte)
+             * 'encoding' => 'utf8mb3' // On MariaDB/MySQL for limited UTF-8 encoding (3-byte, not recommended)
+             * 'encoding' => 'utf8' // On any other RDBMS for full 4-byte UTF-8 encoding (4-byte)
+             */
+            // 'encoding' => 'utf8mb4',
```
dereuromark: This line is not necessary with a an inline comment above.

config/app.default.php
```diff
@@ -232,7 +237,14 @@
             'username' => 'my_app',
             'password' => 'secret',
             'database' => 'my_app',
-            'encoding' => 'utf8',
+            /**
```
dereuromark: This is not a doc block but inline comment, thus `/*`.

config/app.default.php
inoas: Do you agree?

dereuromark: PS: note the missing whitespace after // if it is commented out code vs. comment.

dereuromark: I thought you would keep the current

     'encoding' => 'utf8',

If you indeed remove the line, then you can of course keep the 2nd commented out one :)
But maybe we should keep the current default as it is valid for 3/4 of the systems?

inoas: I don't think so.
- Most starters will use mysql/mariadb
- Those are the most likely to just out-comment the block and have broken defaults by default (broken: 3-byte utf8 but you would not know until you encounter run-time problems)
- For the other (more power-)users: They would either read the docblock and not out-comment it, or outcomment and adapt the block. If not they will face an error right in their face like 'utf8mb4 not supported encoding' on postgres and alike and not just encounter it at some point in future without knowing about it.

config/app.default.php
```diff
@@ -232,7 +239,14 @@
             'username' => 'my_app',
             'password' => 'secret',
             'database' => 'my_app',
-            'encoding' => 'utf8',
+            /*
+             * You do not need to set this flag to use full utf-8 encoding (internal default since CakePHP 3.6).
+             * If you do these are the recommendations:
+             * 'encoding' => 'utf8mb4' // On MariaDB/MySQL for full UTF-8 encoding (4-byte)
+             * 'encoding' => 'utf8mb3' // On MariaDB/MySQL for limited UTF-8 encoding (3-byte, not recommended)
```
ravage84: If it is not recommended, I wouldn't list it under the recommendations :trollface:

inoas: If you want utf8mb3 (aka how it has been before) then that's the recommendation, but yes we can remove it.

inoas:
Anything left to be done?

PR #557, opened 2017-10-17
Replaces deprecated `Response::type()` by `Response::getType()` (#555)

### LINKED ISSUE #555: Response::type() within AppController's beforeRender()
This is a (multiple allowed):

* [ ] bug
* [X] enhancement

* CakePHP Application Skeleton Version: CakePHP 3.5.3

### What I did
I was just looking at the [**3.4 migration guide**](https://book.cakephp.org/3.0/en/appendices/3-4-migration-guide.html#request-response-deprecations)...

### What happened
... when I saw that `Response::type()` had been deprecated and should be replaced by `Response::withType()`.

### What I expected to happen
But when I looked into the new app skeleton provided here, the old `type()` method is still used within the `AppController`'s `beforeRender()` (_JSON / XML_ detection).
Should not this be changed ?

A big thanks 👋

PS : I've dug a bit the WWW, but didn't find a thing about that, am I misunderstanding ?

HorlogeSkynet:
I'm back !

So this is a workaround using `RequestHandler::mapAlias()` and `RequestHandler::responseType()`.
The only problem is the variable `$mimeType` may receive either a `string` or an `array`.
This is why the distinction of cases is done.

I've done some tests on my own, it _should_ work.
BUT, it'd be very interesting if some expert eyes could catch a glimpse of this 👀 

Last point : This is "bigger" than before, and if someone here has a better idea, let it known !

++ 👋

EDIT : 'would fix #555.
EDIT_2 : Sorry for the second commit, didn't know that Travis would complain about that...

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -63,16 +63,16 @@ class AppController extends Controller
         // Note: These defaults are just to get started quickly with development
         // and should not be used in production. You should instead set "_serialize"
         // in each action as required.
-        if(!array_key_exists('_serialize', $this->viewVars)) {
+        if (!array_key_exists('_serialize', $this->viewVars)) {
             $mimeType = $this->RequestHandler->mapAlias(
-                            $this->RequestHandler->responseType());
-            if(!is_array($mimeType)) {
-                if(in_array($mimeType, ['application/json', 'application/xml'])) {
+                $this->RequestHandler->responseType()
+            );
+            if (!is_array($mimeType)) {
+                if (in_array($mimeType, ['application/json', 'application/xml'])) {
                     $this->set('_serialize', true);
                 }
-            }
-            else if(in_array('application/json', $mimeType) ||
-                    in_array('application/xml',  $mimeType)) {
+            } elseif (in_array('application/json', $mimeType) ||
+                      in_array('application/xml', $mimeType)) {
                 $this->set('_serialize', true);
             }
         }
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -55,7 +55,7 @@ class AppController extends Controller
     /**
      * Before render callback.
      *
-     * @param  \Cake\Event\Event $event The beforeRender event.
+     * @param \Cake\Event\Event $event The beforeRender event.
      * @return \Cake\Http\Response|null|void
      */
     public function beforeRender(Event $event)
@@ -63,13 +63,10 @@ class AppController extends Controller
         // Note: These defaults are just to get started quickly with development
         // and should not be used in production.
         // You should instead set "_serialize" in each action as required.
-        if (!array_key_exists('_serialize', $this->viewVars)) {
-            $contentType = $this->response->getHeaderLine('Content-Type');
-            if (strpos($contentType, 'application/json') !== false ||
-                strpos($contentType, 'application/xml') !== false
-            ) {
-                $this->set('_serialize', true);
-            }
+        if (!array_key_exists('_serialize', $this->viewVars) &&
+            in_array($this->response->getType(), ['application/json', 'application/xml'])
+        ) {
+            $this->set('_serialize', true);
         }
     }
 }
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -63,10 +63,18 @@ class AppController extends Controller
         // Note: These defaults are just to get started quickly with development
         // and should not be used in production. You should instead set "_serialize"
         // in each action as required.
-        if (!array_key_exists('_serialize', $this->viewVars) &&
-            in_array($this->response->type(), ['application/json', 'application/xml'])
-        ) {
-            $this->set('_serialize', true);
+        if(!array_key_exists('_serialize', $this->viewVars)) {
+            $mimeType = $this->RequestHandler->mapAlias(
+                            $this->RequestHandler->responseType());
+            if(!is_array($mimeType)) {
+                if(in_array($mimeType, ['application/json', 'application/xml'])) {
+                    $this->set('_serialize', true);
+                }
+            }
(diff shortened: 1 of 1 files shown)


src/Controller/AppController.php
```diff
@@ -63,10 +63,18 @@ public function beforeRender(Event $event)
         // Note: These defaults are just to get started quickly with development
         // and should not be used in production. You should instead set "_serialize"
         // in each action as required.
-        if (!array_key_exists('_serialize', $this->viewVars) &&
-            in_array($this->response->type(), ['application/json', 'application/xml'])
-        ) {
-            $this->set('_serialize', true);
+        if (!array_key_exists('_serialize', $this->viewVars)) {
+            $mimeType = $this->RequestHandler->mapAlias(
```
markstory: I'm not sure we can rely on `RequestHandler` being loaded. In addition to this approach you can also use `$this->response->getHeaderLine('Content-Type')` to read the content type header. That might result in simpler code that lets you continue to use `in_array()`.

HorlogeSkynet: Why couldn't we rely on `RequestHandler` here ?

--> I didn't know about that one, I may simplify this if you're OK with the concept !

inoas: https://gist.github.com/inoas/e58900d3e266a8827b988142af490c6f
... or about the same in class `\Cake\Utility\Text` may help

ADmad:
I am a bit confused, what exactly is the purpose of this change?

saeideng:
>deprecated `Response::type()`
I not found deprecated method
can anyone help to me ?
https://github.com/cakephp/cakephp/blob/fcb0b36eb44460cb5bc50ab6a999b9ab2038ab15/src/Http/Response.php#L1042-L1046

HorlogeSkynet:
As specified during the issue (#555), it's what the [CakePHP 3.4's Migration Guide](https://book.cakephp.org/3.0/en/appendices/3-4-migration-guide.html#request-response-deprecations) announced in the past.

But I don't know either why the source documentation does not specified it. Maybe an oversight ?

markstory:
@HorlogeSkynet The migration notes only document the write side of `type()` being deprecated. With the encoding challeges, we might want to consider adding `getType()`.

HorlogeSkynet:
@markstory This is what is effectively, indeed.
Where are we heading then ?
Do I try something with your `getHeaderLine()`, or do we wait for a proper _getter_ ?

ADmad:
This PR highlights the fact that we are probably missing a method on Response class to get the currently set type/mimetype. There's getMimeType() but it doesn't do what I would intuitively expect. It returns the mapped mimetype for given type instead of currently set mimetype.

HorlogeSkynet:
Well, I couldn't resist, the latest version is simple.
Keep me posted @markstory 👍

markstory:
@ADmad There are higher level abstractions in RequestHandler for setting content types. For lower level APIs there is `withType()`, `withCharset()` and `withHeader()`. What is not 'easy' right now is something like `getType()` which would get the current content-type without the charset bits.

src/Controller/AppController.php
```diff
@@ -55,18 +55,21 @@ public function initialize()
...
-        if (!array_key_exists('_serialize', $this->viewVars) &&
-            in_array($this->response->type(), ['application/json', 'application/xml'])
-        ) {
-            $this->set('_serialize', true);
+        // and should not be used in production.
+        // You should instead set "_serialize" in each action as required.
+        if (!array_key_exists('_serialize', $this->viewVars)) {
+            $contentType = $this->response->getHeaderLine('Content-Type');
+            if (strstr($contentType, 'application/json') ||
+                strstr($contentType, 'application/xml')
```
markstory: These `strstr()` calls could be `strpos()`. It is a bit faster and uses less memory. While it is likely not relevant here it is a good practice to adopt.

HorlogeSkynet: Yeah it's a beautiful confusion, I had planned to user `strpos()` 😢
May I force push ?

markstory: Of course. Force push and rebase as much as you need.

saeideng:
@markstory  I wrote some code/test for Implementing getType()
you can see it in new PR

ADmad:
> What is not 'easy' right now is something like getType() which would get the current content-type without the charset bits.

That's what I was referring to :)

src/Controller/AppController.php
```diff
@@ -55,18 +55,21 @@ public function initialize()
...
+     * @param  \Cake\Event\Event $event The beforeRender event.
      * @return \Cake\Http\Response|null|void
      */
     public function beforeRender(Event $event)
     {
         // Note: These defaults are just to get started quickly with development
-        // and should not be used in production. You should instead set "_serialize"
-        // in each action as required.
-        if (!array_key_exists('_serialize', $this->viewVars) &&
-            in_array($this->response->type(), ['application/json', 'application/xml'])
```
saeideng: new Cakephp released 
you can replace this by `$this->response->getType()` now

HorlogeSkynet: Yeah thank you, I was waiting for the tag. I start working on this ASAP 👌

src/Controller/AppController.php
```diff
@@ -55,18 +55,21 @@ public function initialize()
     /**
      * Before render callback.
      *
-     * @param \Cake\Event\Event $event The beforeRender event.
+     * @param  \Cake\Event\Event $event The beforeRender event.
```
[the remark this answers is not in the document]
HorlogeSkynet: Yeah that was `phpcbf`... Does it matter ?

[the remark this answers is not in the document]
ADmad: If you revert it back it won't :)

Issue #368, opened 2016-04-13
Easier integration with Let's Encrypt


markstory:
_From @watermark on April 13, 2016 17:6_

This is a (multiple allowed):
- [ ] bug
- [x] enhancement
- [ ] feature-discussion (RFC)
- CakePHP Version: Any
- Platform and Target: Apache
### Summary

Let's Encrypt is out of beta now.  I think we can all agree that more HTTPS is a good thing.  I'd like to suggest a modification to CakePHP to better support Let's Encrypt.
### Issue

The automated verification that the letsencrypt client does places a file at "/.well-known/acme-challange/filename", where filename is a name set by the letsencrypt verification process.  CakePHP redirects all requests to the webroot directory, which can interfere with this process.

Technically, it is possible to setup the let's encrypt client to place the challenge file structure in the webroot directory; however, many of the automated utilities that are coming out don't expose this setting.  You might argue that it's a failing of those utilities, but it's also an easy fix in CakePHP.
### Recommendation

Add lines to the top of the default htaccess such that it passes the request on to Apache, bypassing cakephp.

Add "RewriteRule ^(\.well-known/.*)$ $1 [L]" as the first rewrite rule in the root htaccess and the app/.htaccess
### The .well-known directory

The ".well-known" directory is defined in RFC 5785.  Given my understanding of the RFC, I believe it's appropriate to expose the entire directory as suggested above.

_Copied from original issue: cakephp/cakephp#8642_

markstory:
Seems reasonable to me. Could you send a pull request to cakephp/app with the requested changes?

ravage84:
Can we make this rewrite rule to be commented out by default?
The less rules, the faster the processing.

But :+1: for making the Let's Encrypt integration easier, nontheless.
Exposing the whole ``.well-known`` directory seems to be :ok: .

Also, we should back-port this to 2.x as it non-obtrusive, escpecially when commented out.

BassemN:
@watermark :+1: @ravage84 :+1:

watermark:
I'm for making it the default, not commented out by default.

I'd argue that the one extra rule will have a near zero impact on performance.  Especially because Cake is already using an htaccess and enables mod_rewrite.  I cannot find any performance numbers to support either argument.

Just as you're free to remove the htaccess and enable "CakePHP pretty URLs", you would be free to remove this new addition as well.

ravage84:
Created a PR for 2.x
https://github.com/cakephp/cakephp/pull/8647

[...]

PR #558, opened 2017-10-17
Update htaccess to support LetsEncrypt

watermark:
Update rewrite rules in the base htaccess to directly expose the ".well-known" directory.  This adds support for LetsEncrypt utilities to successfully request certificates.

This was merged previously for Cake2, but not Cake3.  Issue #368.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -6,6 +6,7 @@
 
 <IfModule mod_rewrite.c>
     RewriteEngine on
+    RewriteRule    ^(\.well-known/.*)$ $1 [L]
     RewriteRule    ^$    webroot/    [L]
     RewriteRule    (.*) webroot/$1    [L]
 </IfModule>


ADmad:
I have setup SSL using let's encrypt and did not require any change to the default .htaccess. Also in production the document root should be set to `webroot` folder so don't see how making any change to the .htaccess file outside webroot does any good.

watermark:
If you have the document root set to webroot, then this change is not needed, which is likely why your installs didn't need it.  In many installs (shared hosting) you don't have the option of changing the document root.

.htaccess
```diff
@@ -6,6 +6,7 @@
 
 <IfModule mod_rewrite.c>
     RewriteEngine on
+    RewriteRule    ^(\.well-known/.*)$ $1 [L]
```
ravage84: I would prefer this to be commented out by default. Those who need it, can uncomment it.
This file is part of the app and thus becomes userland and thus should be part of a first party version control system.

markstory: If you're using an .htaccess file at all you have already made a performance mistake :smile:

ADmad:
> In many installs (shared hosting) you don't have the option of changing the document root.

Do shared hosting provider support? I personally  don't know of any.

inoas:
Yes they do (DomainFactory at least)

markstory:
> I'd still not add anything to cakephp/app that makes it easier to run things of another doc root than

Why?

inoas:
Because I think we should never encourage bad practice. If someone cannot implement bad practice, because it is too complex to add a line to .htaccess, then they better not have it made easy for them to use non-webroot defaults.

This PR is okay, I just wouldn't put it into the skeleton but into a concrete app / the book if necessary. No worries.

markstory:
Lots of people use shared hosts. I think we can use 1 line of code to make securing their applications easier.

PR #559, opened 2017-10-18
CakePHP 3.5 env() function deprecated …

acatejr:
In the CakePHP 2.5 docs I noticed that env() is deprecated and being replaced by getEnv().  I could be wrong and apologies if I am.  I went through just the config/app.default.php file and replaced all occurances of env() with getEnv().  I hope this was helpful.

**PLEASE NOTE:**

This is only a issue tracker for issues related to the CakePHP Application Skeleton.
For CakePHP Framework issues please use this [issue tracker](https://github.com/cakephp/cakephp/issues).

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.

Always follow the [contribution guidelines](https://github.com/cakephp/cakephp/blob/master/.github/CONTRIBUTING.md) guidelines when submitting a pull request. In particular, make sure existing tests still pass, and add tests for all new behavior. When fixing a bug, you may want to add a test to verify the fix.

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -9,7 +9,7 @@ return [
      * Development Mode:
      * true: Errors and warnings shown.
      */
-    'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),
+    'debug' => filter_var(getEnv('DEBUG', true), FILTER_VALIDATE_BOOLEAN),
 
     /**
      * Configure basic information about the application.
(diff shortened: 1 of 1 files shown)


saeideng:
`env()` is true 
`env()`  is not deprecate
see book https://book.cakephp.org/3.0/en/core-libraries/global-constants-and-functions.html#env

saeideng:
and we have not any global method with naming `getEnv` in core
what is you try to that is `ServerRequest::env()` that is now deprecated 
and you should use `ServerRequest::getEnv()`/ `ServerRequest::withEnv()` instead.

ADmad:
`env()` is not deprecated.

acatejr:
Apologies!!!  I was looking in the wrong class.

PR #563, opened 2017-11-21
RewriteBase / used to be in CakePHP2

inoas:
It is required for some shared hosts and otherwise does no harm in my experience.
(Affects: DomainFactory / HostEurope)

### DIFF
--- a/webroot/.htaccess
+++ b/webroot/.htaccess
@@ -1,5 +1,6 @@
 <IfModule mod_rewrite.c>
     RewriteEngine On
-    RewriteCond %{REQUEST_FILENAME} !-f
-    RewriteRule ^ index.php [L]
+    RewriteBase   /
+    RewriteCond   %{REQUEST_FILENAME} !-f
+    RewriteRule   ^ index.php [L]
 </IfModule>
--- a/webroot/.htaccess
+++ b/webroot/.htaccess
@@ -1,6 +1,7 @@
 <IfModule mod_rewrite.c>
     RewriteEngine On
-    RewriteBase   /
+    # RewriteBase / may be required by some hosts
+    #RewriteBase   /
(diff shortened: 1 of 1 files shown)


webroot/.htaccess
```diff
@@ -1,5 +1,6 @@
 <IfModule mod_rewrite.c>
     RewriteEngine On
-    RewriteCond %{REQUEST_FILENAME} !-f
-    RewriteRule ^ index.php [L]
+    RewriteBase   /
```
ADmad: This should be commented otherwise it will cause problem for sane hostings which don't need it.

webroot/.htaccess
```diff
@@ -1,5 +1,7 @@
 <IfModule mod_rewrite.c>
     RewriteEngine On
-    RewriteCond %{REQUEST_FILENAME} !-f
-    RewriteRule ^ index.php [L]
+    # RewriteBase / may be required by some hosts
+    #RewriteBase   /
```
saeideng: `RewriteBase   /`
Broke my app  in local :)

inoas: Really? Never broke mine neither in cakephp2 nor 3.

BassemN: it brokes my app also

BassemN: I think this is related https://github.com/joomla/joomla-cms/blob/staging/htaccess.txt#L58-L64

ravage84:
@inoas closing now. Feel free to open a new PR with a more verbose comment as "RewriteBase / may be required by some hosts".

inoas:
Well it has always been required by DomainFactory and I have never had problems on any production or dev-box and it has been in CakePHP2 for quite a while (I had PRed that myself years ago).

Never mind through. Easy to add, just a hassle for newbies.

PR #564, opened 2017-11-21
On some shared hosts STDIN is not defined.

inoas:
add STDIN fallback


src/Console/Installer.php
```diff
@@ -14,6 +14,10 @@
  */
 namespace App\Console;
 
+if (defined('STDIN') === false) {
```
ADmad: `if (!defined('STDIN')) {`

PR #565, opened 2017-11-23
Don't insert final newline for twig files.

ADmad:


dereuromark:
I guess we will need this then, yeah :+1:

PR #569, opened 2017-12-04
Remove default serialization of all view vars.

ADmad:
This is an unsafe default and a comment warning is not enough.

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -51,22 +51,4 @@ class AppController extends Controller
         //$this->loadComponent('Security');
         //$this->loadComponent('Csrf');
     }
-
-    /**
-     * Before render callback.
-     *
-     * @param \Cake\Event\Event $event The beforeRender event.
-     * @return \Cake\Http\Response|null|void
-     */
-    public function beforeRender(Event $event)
-    {
-        // Note: These defaults are just to get started quickly with development
-        // and should not be used in production. You should instead set "_serialize"
-        // in each action as required.
-        if (!array_key_exists('_serialize', $this->viewVars) &&
-            in_array($this->response->type(), ['application/json', 'application/xml'])
-        ) {
-            $this->set('_serialize', true);
-        }
-    }
 }


dereuromark:
As alternative, how about putting it into a protected method and then using it commented out?
This way you can activate with a single line commenting in.

markstory:
Refs cakephp/cakephp#11506

chinpei215:
I would suggest to stop loading `RequestHandlerComponent` by default. Even if we remove the problematic code, Cake3 applications would still leak sensitive information, as Bake generates `$this->set('_serialize', ['users']);` or something.

Also, I think it would be better to announce this issue officially, as almost all Cake3 applications would be affected by this issue. Though I have already announce this issue to people in **#japanese** channel on our Slack, I am not sure what is the best way to secure their applications. For now I have recommended to remove `$this->loadCompnent('RequestHandler');` from their `AppController`.

ADmad:
@dereuromark We would still be providing a way to use the framework in an unsafe way so best to remove it.

@chinpei215 I am not in favor removing `RequestHandlerComponent` from being loaded by default. I am pretty sure it's used in most apps. While bake might be generating `$this->set('_serialize', ['users']);` is pretty evident that if you serialize `$users` all user data is gonna end up in the output unless you restrict the fields.

inoas:
Also by default sensitive fields such as passwords are hidden by bakes defaults AFAIR.
But if it must be we can outcomment the `$this->set('_serialize', ['users']);` in bake instead of the RequestHandler?

chinpei215:
> While bake might be generating $this->set('_serialize', ['users']); is pretty evident that if you serialize $users all user data is gonna end up in the output unless you restrict the fields.

I disagree. I think any beginners would not touch the code as they wouldn't understand what it means.

markstory:
Removing RequestHandlerComponent by default will make building API's require a bit more work.  I think making APIs easy to build should be something we enable as it is a requirement for new applications that use client side frameworks.

chinpei215:
Seriously? I don't think adding one line `$this->loadComponent('RequestHandler')` is a hard work.
Imagine if you have a users' profile page on your website. The page doesn't show you any sensitive information, such as `email`, `family_name`, `given_name` , `address`, and so on. It shows people `nickname` only. However, if you access the page with `Accept: application/json` header, you will see those sensitive information. It is a very bad behavior.
When we consider security issue, it's best to be on the safe side. When we compare 'adding a single `$this->loadComponent('RequestHandler')` (otherwise you cannot get json response)' with 'removing `_serialize` from your baked code (otherwise you will expose any sensitive information), the latter wins.
If you still have not changed your opinion, let's ask Mozilla Secure Open Source for their opinions. I think they overlooked this issue. But if they knew this issue, they would have marked this as critical. Because any websites made with Cake3 are possible to expose any sensitive information if the developers don't have enough knowledge about `RequestHandler`. I have never used the feature in 2.x. So if I were made a website with Cake3, my website would have become vulnerable.

inoas:
What about shipping a prefixed ApiAppController with requesthandler and auto serialization enabled?

chinpei215:
I wouldn't oppose it.

markstory:
@chinpei215 Those are fair points, and you're right adding one component load is not hard.

lorenzo:
Maybe we should have a different component (or controller method) to enable automatic json views?

chinpei215:
It could be an another approach.

saeideng:
how happen if set `$this->set('_serialize', false);` here ?

rchavik:
Related #501

markstory:
Backported to `master` in 84ec633f3127cd431696cea5c2cb92756072c53b

PR #578, opened 2018-01-17
Update composer.json

EOM:
This will add more compactness to the CakePHP to work in an AppEngine environment with Flex. It only remains to create an app.yaml as requested by AppEngine and that's it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,8 @@
     "license": "MIT",
     "require": {
         "php": ">=5.6",
+        "ext-intl": "*",
+        "ext-mbstring": "*",
         "cakephp/cakephp": "3.5.*",
         "cakephp/migrations": "^1.0",
         "cakephp/plugin-installer": "^1.0",


dereuromark:
Shouldn't at least intl be optional (if you dont need any of it) and thus be suggest or require-dev?
Making this a hard requirement in the past had causes people some grieve.

dereuromark:
I see intl still in here, though:
https://github.com/cakephp/cakephp/blob/master/composer.json#L32-L33

ADmad:
As pointed out by @dereuromark, intl and mbstring are already added as dependencies of CakePHP core. So why is duplicating then in app skeleton necessary?

https://github.com/cakephp/cakephp/blob/82fc4b783b0411d73565a7e49122dd7e8721f7ed/composer.json#L32

EOM:
Hello,
Because AppEngine PHP Flex, from Google, looks if it has libraries required in the composer.json root, not in the other composer.json like the one that has CakePHP (https://github.com/cakephp/cakephp/blob/master/composer .json # L32-L33)
Of this menar with there is no error and it is simpler to execute the installation in AppEngine.

More information (Alternative option load ext PHP): 
https://cloud.google.com/appengine/docs/flexible/php/runtime#dependencies

markstory:
The new composer requirements are reasonable to me. They will continue to be dependencies of CakePHP in the foreseeable future. Having them here helps make deployment in appengine easier and doesn't add cruft that needs to be removed for other environments.

EOM:
@ADmad  Yes, in AppEngine only intl. 
Ext mbstring is enable for default in AppEngine Flex

dereuromark:
Closing as the main library CakePHP needs to declare this, not the project level.

PR #581, opened 2018-01-30
Update composer.json josegonzalez/dotenv (2.1.0 => 3.2.0)

EOM:
josegonzalez/dotenv (2.1.0 => 3.2.0)


markstory:
We cannot require only 3.x as it drops support for PHP 5. We could do `2.* | 3.*` though.

josegonzalez:
Nope, I re-added it in 3.2!

markstory:
Well then, we should be good.

PR #583, opened 2018-02-19
Set timezone from `.env`

gotoeveryone:
Like `encoding` and `defaultLocale`, change to get the value from `.env`.


markstory:
Thanks!

PR #585, opened 2018-03-10
add route collection cache

steinkel:
Add route collection caching.

I've been doing some more benchmarks locally (php builtin server) and there could be some cases for applications with a small (<5) number of routes where using cache by default will take 0.1-0.2 milliseconds more. This overhead is caused by the initial cache write, object serialization, network, etc.

I think this is not a strong reason to make it optional, as the benefits for non trivial cake applications are *huge*, but I'm open to suggestions. I've left a note in the Application.php class.

Thanks,

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -118,6 +118,20 @@ return [
             'duration' => '+1 years',
             'url' => env('CACHE_CAKEMODEL_URL', null),
         ],
+
+        /**
+         * Configure the cache for routes. The cached routes collection is built the
+         * first time the routes are processed via `config/routes.php`.
+         * Duration will be set to '+2 minutes' in bootstrap.php when debug = true
+         */
+        '_cake_routes_' => [
+            'className' => 'File',
+            'prefix' => 'myapp_cake_routes_',
+            'path' => CACHE,
+            'serialize' => true,
+            'duration' => '+1 years',
+            'url' => env('CACHE_CAKEROUTES_URL', null),
+        ],
     ],
 
     /**
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -122,7 +122,7 @@ return [
         /**
          * Configure the cache for routes. The cached routes collection is built the
          * first time the routes are processed via `config/routes.php`.
-         * Duration will be set to '+2 minutes' in bootstrap.php when debug = true
+         * Duration will be set to '+2 seconds' in bootstrap.php when debug = true
          */
         '_cake_routes_' => [
             'className' => 'File',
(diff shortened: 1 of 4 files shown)


config/bootstrap.php
```diff
@@ -107,6 +107,7 @@
 if (Configure::read('debug')) {
     Configure::write('Cache._cake_model_.duration', '+2 minutes');
     Configure::write('Cache._cake_core_.duration', '+2 minutes');
+    Configure::write('Cache._cake_routes_.duration', '+2 minutes');
```
markstory: Should this be lower? I could see a 2 min duration getting tedious when doing development.  What do you think of 30 seconds?

lorenzo: Is there any value of caching it during development? Considering how annoying it can be sometimes

config/bootstrap.php
```diff
@@ -107,6 +107,8 @@
 if (Configure::read('debug')) {
     Configure::write('Cache._cake_model_.duration', '+2 minutes');
     Configure::write('Cache._cake_core_.duration', '+2 minutes');
+    // disable router cache during development
+    Configure::write('Cache._cake_routes_.duration', '+0 minutes');
```
ADmad: Won't this still unnecessarily write to cache only to be invalidated? A better option could be changing the middleware setup in `Application.php` to
`->add(new RoutingMiddleware($this, Configure::read('debug') ? null : '_cake_routes'));`

config/bootstrap.php
markstory: Duration of 1-5 seconds? might be small enough to not get in the way.

config/bootstrap.php
saeideng: @ADmad there is a problem 
when you are in product mode cache uses 
when you switch to debug mode cache disabled , without clearing 
but when you switch again to product mode , cake uses old cache

steinkel: Yes, also I think it's better to be consistent and write to cache in debug mode too to help the developer to deal with cache configuration.

config/bootstrap.php
steinkel: fixed, would it be a reason too to reduce debug mode cache times for _core_ and _model_ ?

config/app.default.php
```diff
@@ -118,6 +118,20 @@
...
             'url' => env('CACHE_CAKEMODEL_URL', null),
         ],
+
+        /**
+         * Configure the cache for routes. The cached routes collection is built the
+         * first time the routes are processed via `config/routes.php`.
+         * Duration will be set to '+2 minutes' in bootstrap.php when debug = true
+         */
+        '_cake_routes_' => [
+            'className' => 'File',
```
lorenzo: Can you add a comment here?

```
// Pro Tip: install apcu and change the className to Apc for an even faster router
```

PR #587, opened 2018-03-18
prepare 3.next for betas to test.

dereuromark:
3.next is rather outdated at this point.


dereuromark:
I dont think your merge commit was right, though:
https://github.com/cakephp/app/blob/master/composer.json#L9 should not have any 3.6 usage yet, right?

saeideng:
@dereuromark  fixed , sorry

PR #590, opened 2018-04-06
Add function to try to find a CLI version of PHP

### LINKED ISSUE #589: bin/cake assumes the php executable is the CLI SAPI
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Application Skeleton Version: Every release from 3.0 to at least 3.5.1
* Platform and Target: CentOS with cPanel and EasyApache

### What you did
Ran `bin/cake <shell_name>` from the cron.

### What happened
`bootstrap_cli.php` was not included.

`bootstrap_cli.php` should be included.

`bin/cake` does `exec php`, and then `bootstrap.php` tests `$isCli = PHP_SAPI === 'cli';` to decide whether or not to include `bootstrap_cli.php` (among other things). But depending on the particular build options for PHP, it's possible that `php` is the CGI SAPI, and `php-cli` is needed to get the CLI SAPI. (In my case, presumably not an uncommon hosting environment for Cake users, `/usr/local/bin/php` is CLI, but `/usr/bin/php` is CGI, and the cron's PATH means that the latter is used there. `/usr/bin/php-cli` is a symlink to `/usr/local/bin/php`.)

I was able to fix this (once I eventually tracked down the problem) by changing `bin/cake` to `exec php-cli`, but it seems like it could be nice to build in a check for this. The output of `php -v` could be used, for example, or `echo "<?= PHP_SAPI ?>" | php` or the like.

Zuluru:
Fixes #589.

I have included a way for people to set their required PHP via environment variable, if their setup doesn't match one of the pre-defined options. Is there any practical way that this could be exploited? If so, is there a way to sanitize that input before using it?

### DIFF
--- a/bin/cake
+++ b/bin/cake
@@ -33,14 +33,36 @@ canonicalize() {
     echo "$NAME"
 }
 
+# Find a CLI version of PHP
+findCliPhp() {
+    for TESTEXEC in php php-cli /usr/local/bin/php
+    do
+        SAPI=`echo "<?= PHP_SAPI ?>" | $TESTEXEC 2>/dev/null`
+        if [ "$SAPI" = "cli" ]
+        then
+            echo $TESTEXEC
+            return
+        fi
+    done
+    echo "Failed to find a CLI version of PHP; falling back to system standard php executable" >&2
+    echo "php";
+}
+
 CONSOLE=$(dirname -- "$(canonicalize "$0")")
 APP=$(dirname "$CONSOLE")
 
+# If your CLI PHP is somewhere that this doesn't find, you can define a PHP environment
+# variable with the correct path in it.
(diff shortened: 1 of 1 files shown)


bin/cake
```diff
@@ -33,14 +33,36 @@ canonicalize() {
...
+    echo "Failed to find a CLI version of PHP; falling back to system standard php executable" >&2
+    echo "php";
+}
+
 CONSOLE=$(dirname -- "$(canonicalize "$0")")
 APP=$(dirname "$CONSOLE")
 
+# If your CLI PHP is somewhere that this doesn't find, you can define a PHP environment
+# variable with the correct path in it.
+if [ -z "$PHP" ]
```
ADmad: :+1:

PR #597, opened 2018-04-21
adds method bootstrap to App\Application

Schlaefer:
Activating a plugin with `bin/cake plugin load myPlugin` throws "Your Application class does not have a bootstrap() method. Please add one."

Instead of an error message it would be nice to have the method already in place.


markstory:
Thanks!

PR #602, opened 2018-04-30
Consistent comment blocks

davidyell:
DocComments start with two asterisks.
Ref: http://docs.phpdoc.org/guides/docblocks.html

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -13,12 +13,12 @@
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
 
-/*
+/**
  * Configure paths required to find CakePHP + general filepath constants
  */
 require __DIR__ . '/paths.php';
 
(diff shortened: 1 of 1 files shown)


dereuromark:
No, what you did is wrong AFAIK and has been changed to this on purpose.

Doc blocks != Comments
Doc blocks only exist on functions, methods or classes etc.
All others are just plain and simple comments, and thus do not have the extra *.
This makes especially a difference in tools like phpcs which treat it differently then based on the different token parsing internally.

dereuromark:
Read the red notice there:

> Important A DocBlock is always associated with one, and just one, Structural Element in PHP; so this may either be a file, class, interface, trait, function, constant, method, property or variable.

In your case these simple lines and their comments just make it a comment. Nothing else.

davidyell:
Oh right.

PR #604, opened 2018-05-03
Don't recommend loading deprecated component.

ADmad:


jeremyharris:
Should we add a commented `CsrfProtectionMiddleware` in the middleware stack with a similar comment?

ADmad:
Probably.

Issue #593, opened 2018-04-20
Plugin routes config files are included twice


Schlaefer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Application Skeleton Version: 3.6.0
* Platform and Target: CakePHP 3.6.0

### What you did

register a Plugin with routes in bootstrap.php

```php
Plugin::loadAll([
    'Foo' => ['routes' => true],
]);
```

### What happened

The routes file `plugins/Foo/config/routes.php` is included twice.

### What you expected to happen

The routes file should only be included once.

I assume the call to `Plugins::routes()` in `config/routes.php` isn't necessary in Cake 3.6 anymore?

markstory:
This was a problem in an one of the rc releases, but that issue was fixed. What are the stack traces for each include? You can use `debug(stackTrace())` to output a stack trace.

markstory:
Oh I hadn't considered usage of `Plugin::routes()` being called in the application routes.php file. We should be able to remove that methid call now.

josbeir:
I found this out the hard way, (exceptions about named routes already set). Removing Plugin::routes fixed it. Guess this should be mentioned somewhere in the docs and the app skeleton should be adjusted accordingly when releasing the new version as this is mostly the starting point for most new projects...

Also something to take note of is that the current implementation of **RouterShell is unaware of any routes loaded by the RoutingMiddleware** (which is of course logical but i thought it should mention this).

markstory:
> the current implementation of RouterShell is unaware of any routes loaded by the RoutingMiddleware (which is of course logical but i thought it should mention this) 

I can look into fixing this. I'm pretty sure we can figure out a clever solution to this.

markstory:
The issue with `bin/cake routes` is captured in cakephp/cakephp#12030

PR #610, opened 2018-05-25
Remove Plugin::routes()

markstory:
When using the new plugin system this method is no longer necessary as
the plugin classes include routes during application startup.

Refs #593


markstory:
The plugins don't need to use the new plugin classes.  As long as the app uses `Application` plugin classes will be created dynamically if they are not provided by the plugins.

PR #614, opened 2018-06-20
Use fully qualified class names in config

garas:
Use fully qualified class names everywhere in app.php so it is consistent with other 'className' uses in this file.

This also allow to Ctrl-click in editor (at least in PhpStorm) to open class file and see its config, implementation and easier discover alternate cache engines / mail transports in same directory.

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -88,7 +88,7 @@ return [
      */
     'Cache' => [
         'default' => [
-            'className' => 'File',
+            'className' => 'Cake\Cache\Engine\FileEngine',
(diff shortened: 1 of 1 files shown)


config/app.default.php
```diff
@@ -88,7 +88,7 @@
      */
     'Cache' => [
         'default' => [
-            'className' => 'File',
+            'className' => 'Cake\Cache\Engine\FileEngine',
```
ADmad: Even better would be using `FileEngine::class` instead of string. Same for other classes.

Issue #540, opened 2017-08-21
Should Csrf be removed from AppController and placed in Application instead?


MCF:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Application Skeleton Version: 9eae2648da9dae7511f4ca9070a2cce1114b6ddd

Just wondering if the app should be updated to replace the CsrfComponent (commented out) in the AppController with the new CSRF middleware in Application instead?  Or am I misunderstanding the 3.5 upgrade guide?

markstory:
@MCF CsrfMiddleware should likely be applied as a scoped middleware. This is something that was missed in preparation for 3.5 though.

MCF:
Thanks @markstory, I noted the deprecation of the component in the upgrade guide but was having trouble finding documentation on how the CsrfMIddleware works.  It looks like it is pretty new and the documentation is still pretty thin.  Just checking out the app in case there was any additional guidance and I thought I would mention that it needed to be updated.

markstory:
The existing documentation [can be found here](https://book.cakephp.org/3.0/en/controllers/middleware.html#csrf-middleware) what did you find missing/unclear as I'd like to improve the docs if possible.

MCF:
I was trying to understand how to turn off Csrf for the occasional action.  In complicated interactive forms it can be difficult to have it work with Csrf (Component or Middleware).

It would also be nice if there was some documentation about how Csrf can be made to work with these kinds of forms that are modified via javascript, rather than being disabled entirely in those cases.  For example as a form is filled out items may be enabled/disabled or added/removed depending on choices made earlier in the form.

markstory:
> I was trying to understand how to turn off Csrf for the occasional action. In complicated interactive forms it can be difficult to have it work with Csrf (Component or Middleware).

You would need to use routing scopes to only enable it for certain parts of your application.

> It would also be nice if there was some documentation about how Csrf can be made to work with these kinds of forms that are modified via javascript, rather than being disabled entirely in those cases. 

Wouldn't the CSRF token be sent along with the modified form? Are you thinking of SecurityComponent (which prevents form tampering)?

MCF:
Yes, you are right.  I am conflatting the Security component and the Csrf components a bit here.  Here is a sample snippet of code from a beforeFilter in one of my controllers:

```PHP
        $action = $this->request->getParam('action');
        if (in_array($action, ['apply', 'remove'])) {
[...]

PR #616, opened 2018-06-26
Adding CSRF Protection middleware into Application.php

LubosRemplik:
related to #540

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -61,7 +61,9 @@ class Application extends BaseApplication
             ->add(new RoutingMiddleware($this, '_cake_routes_'))
 
             // Add csrf middleware.
-            ->add(new CsrfProtectionMiddleware());
+            ->add(new CsrfProtectionMiddleware([
+                'httpOnly' => true
+            ]));
 
         return $middlewareQueue;
     }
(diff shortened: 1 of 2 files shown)


HorlogeSkynet:
Hello !

Should we enable `httpOnly` by default ? 🤔

Ciao 👋

LubosRemplik:
Not sure, if yes I would change default options here https://github.com/cakephp/cakephp/blob/master/src/Http/Middleware/CsrfProtectionMiddleware.php#L52

HorlogeSkynet:
I would definitely too ! Does anyone know why this choice has been particularly made ?
I mean, for "standard" apps, this could and should be set to `true` without any problem, am I wrong ? :thinking:

markstory:
The default is `false` to retain backwards compatibility with the CsrfComponent. We cannot change the default value until 4.x. If you'd like to change it please open a pull request targetting `4.x`.

LubosRemplik:
But still we can add the middleware into the Application.php with default settings, can't we?

markstory:
While the default httpOnly value is false, it could be set to `true` in the application template.

LubosRemplik:
Thanks @ravage84 and @markstory for suggestions, both added

PR #617, opened 2018-06-27
move loading plugins into Application

saeideng:
move loading plugins into Application
after releasing cake 3.6.6

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -207,11 +207,3 @@ Type::build('timestamp')
  * Plugin::load('Migrations'); //Loads a single plugin named Migrations
  *
  */
-
-/*
- * Only try to load DebugKit in development mode
- * Debug Kit should not be installed on a production system
- */
-if (Configure::read('debug')) {
-    Plugin::load('DebugKit', ['bootstrap' => true]);
-}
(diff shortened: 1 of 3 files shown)


config/bootstrap.php
```diff
@@ -207,11 +207,3 @@
  * Plugin::load('Migrations'); //Loads a single plugin named Migrations
```
ADmad: Shouldn't this comment also be removed?

src/Application.php
```diff
@@ -61,4 +61,33 @@ public function middleware($middlewareQueue)
...
+
+            $this->addPlugin('Migrations');
+        }
+
+        /*
+         * Only try to load DebugKit in development mode
+         * Debug Kit should not be installed on a production system
+         */
+        if (Configure::read('debug')) {
+            $this->addPlugin(\DebugKit\Plugin::class, ['bootstrap' => true]);
```
ADmad: Specifying bootstrap => true shouldn't be necessary.

config/bootstrap.php
saeideng: whole comments or just `Plugin::load('Migrations...` ?

saeideng:
looks we should wait for releasing 3.6.7
I tested tests with https://github.com/cakephp/cakephp/pull/12300 
it was OK with ^^ patch

PR #620, opened 2018-07-10
Update Application.php

highstrike:
added missing import


dereuromark:
Seems like there is a test missing.

highstrike:
added missing test

PR #626, opened 2018-08-28
Had issues with Asset.timestamp

inoas:
... maybe something like this could be useful?

### DIFF
--- a/config/paths.php
+++ b/config/paths.php
@@ -48,6 +48,8 @@ define('CONFIG', ROOT . DS . 'config' . DS);
 
 /**
  * File path to the webroot directory.
+ * To derive your webroot from your webserver change this to:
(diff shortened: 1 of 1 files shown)


config/paths.php
```diff
@@ -48,6 +48,8 @@
 
 /**
  * File path to the webroot directory.
+ * To derive your webroot from your webserver change this to:
+ * `define('WWW_ROOT', rtrim($_SERVER['DOCUMENT_ROOT'], DS) . DS);`
```
markstory: When would this be necessary?

inoas: The rtrim is because I do not trust the webserver to have the trailing slash.
However the DOCUMENT_ROOT was necessary when I create a cakephp project that had multiple webroots (portalling/sharding) but shares the same application structure.

ADmad: I don't think we need to add comments for such edge case usages. It would unnecessarily confuse users.

PR #630, opened 2018-09-15
Register scoped middleware in Application::routes()

garas:
Recent change to add CSRF middleware to application middleware queue made many users confused due to CSRF token errors.

CSRF, encrypted cookies, authentication etc middlewares are better uses as scoped middlewares.

I suggest to use `Application::routes()` to register (and group) scoped middlewares. This way all middlewares are still configured in single file, some in `routes()`, some in `middleware()`, but separated.

Add `applyMiddleware()` and short example about separate scopes in `routes.php`.

This PR separates app-wide and scoped middlewares, it might be easier to communicate about differences and enabling/disabling scoped middlewares. In my opinion, `routes.php` has more visibility than `Application.php` between beginner CakePHP developers.

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -48,8 +48,8 @@ Router::defaultRouteClass(DashedRoute::class);
 
 Router::scope('/', function (RouteBuilder $routes) {
     /**
-     * Apply a middlewares to the current route scope.
-     * Requires middlewares to be registered via `Application::routes()` with `registerMiddleware()`
+     * Apply a middleware to the current route scope.
+     * Requires middleware to be registered via `Application::routes()` with `registerMiddleware()`
      */
     $routes->applyMiddleware('csrf');
 
@@ -85,7 +85,7 @@ Router::scope('/', function (RouteBuilder $routes) {
 });
 
 /**
- * If you need different set of middlewares or none at all, open new scope and define routes there
+ * If you need different set of middleware or none at all, open new scope and define routes there
  *
  * ```
  * Router::scope('/api', function (RouteBuilder $routes) {
--- a/config/routes.php
+++ b/config/routes.php
@@ -48,6 +48,12 @@ Router::defaultRouteClass(DashedRoute::class);
 
 Router::scope('/', function (RouteBuilder $routes) {
     /**
+     * Apply a middlewares to the current route scope.
+     * Requires middlewares to be registered via `Application::routes()` with `registerMiddleware()`
+     */
+    $routes->applyMiddleware('csrf');
+
+    /**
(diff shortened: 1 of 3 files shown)


src/Application.php
```diff
@@ -57,6 +57,24 @@ public function bootstrap()
...
+     *
+     * @param \Cake\Routing\RouteBuilder $routes A route builder to add routes into.
+     * @return void
+     */
+    public function routes($routes)
+    {
+        // Register scoped middlewares for use in routes.php
+        $routes->registerMiddleware('csrf', new CsrfProtectionMiddleware([
+            'httpOnly' => true
+        ]));
```
markstory: You can register this middleware in routes.php as well if you want everything in one place.

garas: Some middlewares requires `Application` as `$this`, some like `SecurityHeaderMiddleware` or `BodyParserMiddleware` have somewhat complex configuration or logic and I think `config/` should avoid complex logic, code/logic should be in `src/`.

`routes.php` should be for configuring routes, not middlewares.

markstory: Ok. The parent method is just a require for `config/routes.php` and the `$routes` variable is available there too.

config/routes.php
```diff
@@ -47,6 +47,12 @@
 Router::defaultRouteClass(DashedRoute::class);
 
 Router::scope('/', function (RouteBuilder $routes) {
+    /**
+     * Apply a middlewares to the current route scope.
```
ravage84: middlewares  -> middleware

config/routes.php
```diff
@@ -47,6 +47,12 @@
...
+     * Requires middlewares to be registered via `Application::routes()` with `registerMiddleware()`
```
ravage84: Plural of middleware is middleware. Confusing, but it is...

config/routes.php
```diff
@@ -77,3 +83,14 @@
      */
     $routes->fallbacks(DashedRoute::class);
 });
+
+/**
+ * If you need different set of middlewares or none at all, open new scope and define routes there
```
ravage84: middlewares -> middleware

src/Application.php
```diff
@@ -57,6 +57,24 @@ public function bootstrap()
...
+     * Define the routes for an application.
+     *
+     * Use the provided RouteBuilder to define an application's routing, register scoped middlewares.
+     *
+     * @param \Cake\Routing\RouteBuilder $routes A route builder to add routes into.
+     * @return void
+     */
+    public function routes($routes)
+    {
+        // Register scoped middlewares for use in routes.php
```
ravage84: middlewares  -> middleware

PR #632, opened 2018-09-18
Sentence structure

groovenectar:
Using one form of "interact" in the first sentence. Changed "experiment and explore" to only "explore," partly to line the columns, and with the thought that one could perhaps imply the other.


markstory:
Thank you 👍

PR #636, opened 2018-09-28
Some more help on when to use .env files

lorenzo:
Added a note on when using .env files is not OK

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,13 +9,13 @@
         "cakephp/cakephp": "3.6.*",
         "cakephp/migrations": "^2.0.0",
         "cakephp/plugin-installer": "^1.0",
-        "josegonzalez/dotenv": "3.*",
         "mobiledetect/mobiledetectlib": "2.*"
     },
     "require-dev": {
         "cakephp/bake": "^1.1",
         "cakephp/cakephp-codesniffer": "^3.0",
         "cakephp/debug_kit": "^3.15.0",
+        "josegonzalez/dotenv": "3.*",
         "psy/psysh": "@stable"
     },
     "suggest": {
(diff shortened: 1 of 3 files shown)


bravo-kernel:
aha 💡 👍

config/bootstrap.php
```diff
@@ -50,7 +50,7 @@
  * variables as required.
  *
  * It is HIGHLY discouraged to use a .env file in production, due to security risks
```
bravo-kernel: **an** .env file ?

lorenzo: a `dot env`... that's at least how I read it

bravo-kernel: That's totally correct then and... a very interesting moment of insight into my perception of things  👍

composer.json
```diff
@@ -9,13 +9,13 @@
...
         "cakephp/migrations": "^2.0.0",
         "cakephp/plugin-installer": "^1.0",
-        "josegonzalez/dotenv": "3.*",
         "mobiledetect/mobiledetectlib": "2.*"
     },
     "require-dev": {
         "cakephp/bake": "^1.1",
         "cakephp/cakephp-codesniffer": "^3.0",
         "cakephp/debug_kit": "^3.15.0",
+        "josegonzalez/dotenv": "3.*",
```
ADmad: Nice moving this to `require-dev`.

dereuromark: Wouldnt it then also have to be in suggest for clarification?

dereuromark: OK, I guess this is fine :)

config/bootstrap.php
AD7six: could also have cheated `use .env fileS` :sunglasses:

bravo-kernel: 👏

PR #641, opened 2018-10-17
Pass middleware config through constructor.

ADmad:
Explicitly passing the config allows user to better understand how the error config is used and they could conditionally modify it if needed.

In 4.x I would like remove direct use of `Configure::read('Error')` in `ErrorHandlerMiddleware::__construct()`.


ADmad:
I messed up the constructor args.

PR #642, opened 2018-10-17
Pass middleware config through constructor.

ADmad:
Explicitly passing the config allows user to better understand how the error config is used and they could conditionally modify it if needed.

In 4.x I would like remove direct use of `Configure::read('Error')` in `ErrorHandlerMiddleware::__construct()`.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -86,7 +86,7 @@ class Application extends BaseApplication
         $middlewareQueue
             // Catch any exceptions in the lower layers,
             // and make an error page/response
-            ->add(ErrorHandlerMiddleware::class)
+            ->add(new ErrorHandlerMiddleware(null, Configure::read('Error')))
 
             // Handle plugin/theme assets like CakePHP normally does.
             ->add(new AssetMiddleware([


src/Application.php
```diff
@@ -86,7 +86,7 @@ public function middleware($middlewareQueue)
         $middlewareQueue
             // Catch any exceptions in the lower layers,
             // and make an error page/response
-            ->add(ErrorHandlerMiddleware::class)
+            ->add(new ErrorHandlerMiddleware(null, Configure::read('Error')))
```
markstory: This null is a bit awkward, but hard to remove now :cry:

ADmad: Yeah I don't like it either. Only way to avoid it is drop the first argument. Exception renderer can be set through config too.

saeideng:
@fr3nch13  you are using `cakephp/app` repo as required repo/package 
but you dont use it's `config/app.php` in your test 
and you only use some config in your `tests/bootstrap.php` file that has not  this `Error` key 
you can add this key to `tests/bootstrap.php` or use load `config/app.php` correctly

PR #643, opened 2018-10-24
Remove unused use statement

narendravaghela:
It simply removes on unused `use` statement in `config/routes.php`


markstory:
Thank you 👍

PR #644, opened 2018-10-25
Few more unused use statement in config directory

narendravaghela:
It simply removes on unused use statement in 
- config/bootstrap.php
- config/bootstrap_cli.php

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -31,7 +31,6 @@ require CORE_PATH . 'config' . DS . 'bootstrap.php';
 
 use Cake\Cache\Cache;
 use Cake\Console\ConsoleErrorHandler;
-use Cake\Core\App;
 use Cake\Core\Configure;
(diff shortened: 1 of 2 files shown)


config/bootstrap.php
```diff
@@ -41,7 +40,6 @@
 use Cake\Http\ServerRequest;
 use Cake\Log\Log;
 use Cake\Mailer\Email;
-use Cake\Utility\Inflector;
```
saeideng: for these lines 
https://github.com/cakephp/app/blob/aeca5e2a4c69999beaa743cd41286407fe3b37e3/config/bootstrap.php#L200-L203
we can not remove it

narendravaghela: Oh yea, thank you for the notice.
Actually, that code was commented, so let me add that back.

narendravaghela: Added the statement back to file.

saeideng:
Thank you :+1:

PR #655, opened 2018-12-25
4.x fixes

ADmad:


ADmad:
Not sure why `phpcs` is not found on travis, it is added as dev dependency. https://travis-ci.org/cakephp/app/jobs/472042671#L456

PR #657, opened 2019-01-06
Use protected method instead of too long method and nesting

dereuromark:
what do you think?
after all we used to have bootstrap_cli etc.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -93,5 +93,7 @@ class Application extends BaseApplication
         }
 
         $this->addPlugin('Migrations');
-    }    
+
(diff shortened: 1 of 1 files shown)


src/Application.php
```diff
@@ -37,15 +37,7 @@ public function bootstrap()
         // Call parent to load bootstrap from files.
         parent::bootstrap();
 
-        if (PHP_SAPI === 'cli') {
```
ADmad: I would prefer if we retain the SAPI check here and avoid method call when not needed.

PR #658, opened 2019-01-09
Set the overwrite option to `true` in the (commented) Dotenv initialize

groovenectar:
See here: https://github.com/josegonzalez/php-dotenv

Making this the default should

1) Be harmless, and
2) Potentially save people some headaches

If you are using a `.env` pattern. Try calling `bin/cake` from within CakePHP itself (via an `exec` of some kind).

It throws `LogicException` because by default it doesn't want to overwrite the variables when it reads the `.env` file again.

The underlying issue is hard to spot when you are running into it, but it is this.

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -52,9 +52,9 @@ use Cake\Utility\Security;
 // if (!env('APP_NAME') && file_exists(CONFIG . '.env')) {
 //     $dotenv = new \josegonzalez\Dotenv\Loader([CONFIG . '.env']);
 //     $dotenv->parse()
-//         ->putenv()
-//         ->toEnv()
-//         ->toServer();
+//         ->putenv(true)
+//         ->toEnv(true)
+//         ->toServer(true);
 // }
 
 /*


markstory:
Calling `bin/cake` is not something that the app skeleton does though. What kind of use cases do you have for doing that where you couldn't use more robust sub-process handling?

groovenectar:
@markstory...

For example, if you have in a controller:

`exec('bin/cake')` to call a subroutine of some kind

It will likely throw an error if you are using the default `.env` configuration that ships commented

groovenectar:
So the following things have to be true to run into this:

1) Must be using the Dotenv pattern (which is in use out there, per CakePHP Slack channel)

2) Must try calling a shell routine defined in the same application. E.g., perhaps a manual run of a cron task

markstory:
I don't think there are any existing tests around .env file loading as it is an optional part of the application skeleton. What kind of tests did you have in mind?

josegonzalez:
@groovenectar Hi! I'm the author of php-dotenv. Some general thoughts:

- I think calling `exec('bin/cake some-shell ...')` via web is generally bad practice. Background tasks tend to be long-running and therefore will timeout in odd ways when executed via web.
- I'm a big fan of background jobs - using [cakephp-queuesadilla](https://github.com/josegonzalez/cakephp-queuesadilla). I need to get back on this to add RabbitMQ support, but otherwise its a good way to shuttle jobs into the background for execution.
- Logic can certainly be shared between a background and web process by adding a new class - I generally create `src/Service` and add classes there.
- `.env` loading in the app skeleton definitely needs tests - I know the code works, but more tests == better :)
- While it's terrible practice, I think it's maybe worth adding a ShellDispatcherTrait @markstory. We can document it with a ton of caveats around how its a bad idea, and let developers do `exec` in a saner manner. Thoughts?

groovenectar:
@josegonzalez 

Thanks for your input! I was thinking, sometimes there is a simple CLI task that would go into a cron task. It seems perfectly reasonable to call it via `exec()` in many cases? For example, a "clear cache" task. They can be run in either the foreground or background. Some CLI tasks are really quick.

I can agree that RabbitMQ or Gearman etc are way more appropriate in many cases. But people are hating on the general idea of calling CLI tasks with `exec()`... which I don't think is reasonable, as different cases call for different measures, not to mention time constraints...

The issue comes up when the presumably the same `.env` file is called with the same variable names.

So anyone who doesn't use the overwrite option would get the error, and their code wouldn' t work until they update something.

Here is a weird edge case perhaps: CakePHP reads its `.env` file, and has to call an external process -- and let's please just presume that calling an external process is possible, practical, and ideal in this case.

But say that the external process is reading a different `.env` file, but it still has one or more variables with the same name... I wonder:

1) If it would still throw the error, or

2) When overwrite is set to `true`, as I am suggesting, if the other process would then unexpectedly take on the variable values of the CakePHP app, or vice-versa. This would be a generally undesirable effect of having overwrite mode as the default.

So there is a potential test I can try when there is time.

In the case that I'm talking about... The CLI process is part of the same CakePHP app, and would be reading from the same `.env` file. So thinking from that angle, using overwrite seemed completely harmless.

Perhaps, a decent solution could be to add something in the comment itself about overwrite mode. I might try to come up with something. But just wanted to share these findings and the solution I came up with to this case.

josegonzalez:
I think calling `exec` is probably fine - I wouldn't do it via the web, but here we are. I think doing that to run CakePHP code is less okay, and we should add a ShellDispatcherTrait or something to make that something like this possible:

```
$this->dispatchShell('shell', 'task', $args...)
```

Alternatively, you _should_ be creating a class you can invoke from either the shell or web context, and move your logic into that. That way its just regular code you're invoking, and not some subprocess.

groovenectar:
`ShellDispatcherTrait` sounds neat... I like the idea of having one point of entry for certain processes / tasks, and that would still accomplish that....

Going with your insight as the author of php-dotenv, I don't know that this PR is worth merging. I think that with a certain pattern it is necessary, and with *most* patterns, or any that I know of, it should be harmless... If people do run into, it may be hard to spot, and this would preemptively remedy it.

Found a reference to it here: https://github.com/cakephp/cakephp/issues/12549#issuecomment-420572448

But no one's really reported it on GitHub... and I guess I wouldn't have if I hadn't made the PR after finding what was going on...

It is sounding like it should be left as-is

josegonzalez:
I think you're right. Probably our best bet is maybe a section in the shell docs that says not to do this and instead refactor the code such that you can call a service class instead?

groovenectar:
`ShellDispatcherTrait` seems like too much effort for something that can be accomplished in another way. Although CakePHP is structured in a way that could make it feasible and even an interesting challenge. Perhaps, it can just be let alone. If someone runs into it on the issues tracker, feel free to ping me if you remember and I'll give the rundown from the way I understand it

PR #660, opened 2019-01-28
add html5 video binary setting, webp

inoas:


.gitattributes
```diff
@@ -1,5 +1,5 @@
 # Define the line ending behavior of the different file extensions
-# Set default behavior, in case users don't have core.autocrlf set.
+# Set default behaviour, in case users don't have core.autocrlf set.
```
garas: Look one line above too

dereuromark: American English always for documentation. So behavior and color etc is correct.

markstory: I flipped it back after merge.

PR #664, opened 2019-03-23
Add infos for Postgres

inoas:
Migration from MySQL/MariaDB to Postgres was hard cause of:  `Cannot describe users. It has 0 columns. `
Which is because there was no schema configured.

@lorenzo could/should we make the Datasources.default.schema key readOrFail for the postgres driver?

Aside that issue, these cost some time: Related/Off Topic: For migration from MariaDB to Postgres I collected these infos:
- Move from Auto Increment Integer to UUIDs, also the table class validators needs to be adapted to validate uuid intead of nonNegativInteger for PKs and FKs: https://gist.github.com/inoas/8738e94474ffdb5c568d263ae8601d92
- Export MariaDB, Import Postgres - note: no CURRENT_DATETIME support and dropping MariaDB enums was a good idea https://gist.github.com/inoas/cb7fee79cdde22f265cfa24d478dd97d

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -250,9 +250,9 @@ return [
             'persistent' => false,
             'host' => 'localhost',
             /*
-             * CakePHP will use the default DB port based on the driver selected
-             * MySQL on MAMP uses port 8889, MAMP users will want to uncomment
-             * the following line and set the port accordingly
+             * CakePHP will use the default DB port based on the driver selected.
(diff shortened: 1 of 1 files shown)


markstory:
Tips on moving data from Maria to Postgres sounds like something that can exist outside of the CakePHP documentation. It would make a great blog post I bet :smile:

PR #665, opened 2019-03-27
PHPDoc | Missing @throws tag

AdrienLZ:
Correct PHPDoc warning about missing @throws tag

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -35,6 +35,7 @@ class AppController extends Controller
      *
      * e.g. `$this->loadComponent('Security');`
      *
+     * @throws \Exception
      * @return void
      */
     public function initialize()


ADmad:
`initialize()` doesn't throw any exception directly. There could be an exception bubbling up from one of the method calls made inside it but would be try for pretty much every method throughout the code.

AdrienLZ:
> `initialize()` doesn't throw any exception directly. There could be an exception bubbling up from one of the method calls made inside it but that would be true for pretty much every method throughout the code.

It complain on **loadComponent**.

dereuromark:
Exactly, you should never document non first level exceptions here as this cannot be automated or enforced, leading to copy paste issues and lies.

dereuromark:
In your IDE? Fix you config there, this is a developer issue.

dereuromark:
E.g. In phpstorm: 0 level only.

AdrienLZ:
Thanks @ADmad  @dereuromark  for the explanation.

PR #671, opened 2019-04-22
Turn off routes caching by default.

ADmad:


josegonzalez:
Why?

ADmad:
- Unless you have hundreds of routes connected turning on routes caching provides no real benefit.
- It has limitations which causes issues like this cakephp/cakephp#12562 and cakephp/cakephp#12996. Also you just can't use routes caching if setting routing based middleware which involve use of closures/unseriable code.  Sometimes the unserializable code comes from external libs.
- Recently routes for DebugKit were updated and DeugKit bar stopped showing for everyone who had routes caching enabled. So routes caching shouldn't be enabled when debug is on.

josegonzalez:
Makes sense - might be worth adding such a comment to the commit history so its easier to figure out why a change was made in the future :)

Thanks for explaining :)

ADmad:
Still trying to push past my laziness to write descriptive commit messages :stuck_out_tongue:.

Issue #674, opened 2019-05-07
Default app.php Cache configuration String


delamux:
* [x] feature-discussion (RFC)

Why use in: https://github.com/cakephp/app/blame/master/config/app.default.php#L104
Magic String  fully qualified name, is not better do:

```php
use Cake\Cache\Engine\FileEngine;

return [
//....
    'Cache' => [
        'default' => [
            'className' => FileEngine::class, //versus 'Cake\Cache\Engine\FileEngine',
            'path' => CACHE,
            'url' => env('CACHE_DEFAULT_URL', null),
      ],
//....
]
```

ADmad:
Using the `::class` magic constant is indeed better.

Would you like to submit a patch with suggested changes?

delamux:
> Using the `::class` magic constant is indeed better.
> 
> Would you like to submit a patch with suggested changes?
Yes, I can do it!

PR #677, opened 2019-05-08
#674 replace all class constants

delamux:
Change all String for Constants
https://github.com/cakephp/app/pull/675#issuecomment-490457619


markstory:
Thank you :+1:

PR #678, opened 2019-05-09
"config/requirements.php" Update to PHP 7.2

saeideng:


config/requirements.php
```diff
@@ -20,8 +20,8 @@
 /*
  * You can remove this if you are confident that your PHP version is sufficient.
  */
-if (version_compare(PHP_VERSION, '5.6.0') < 0) {
-    trigger_error('Your PHP version must be equal or higher than 5.6.0 to use CakePHP.' . PHP_EOL, E_USER_ERROR);
+if (version_compare(PHP_VERSION, '7.2.0') < 0) {
```
dereuromark: I wonder if we could read this from the composer.json to have a single source of truth.
Also, why are we still needing this if we have composer and the protection of install here?
Seems redundant either way. If you deploy via copy instead of proper ways, then one shouldnt need to fix up after you.

markstory: This historically has been for the zipball install scenario.

dereuromark: I recommend using Configure::version() here, the overhead is minimal and really quite the edge case IMO.

dereuromark:
The CS fails are unrelated.

PR #679, opened 2019-05-09
Fix up controllers.

dereuromark:


src/Controller/PagesController.php
```diff
@@ -66,5 +69,7 @@ public function display(...$path)
             }
             throw new NotFoundException();
         }
+
+        return null;
```
markstory: Should we `return $this->render()` instead?

ADmad: Yeah, if we are going ot add return typehint it's better to explicitly render and return response, else don't add the typehint.

dereuromark: OK, I adjusted it.

PR #680, opened 2019-05-09
Make phpstan easier to use locally.

dereuromark:
and consolidate with the way scripts are in core (4.x branch).

also speed up travis: from 2 min to 1 min now per run.

I didnt add psalm yet here, this could be done at a later point.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,6 +44,8 @@
         ],
         "cs-check": "phpcs --colors -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP src/ tests/",
         "cs-fix": "phpcbf --colors --standard=vendor/cakephp/cakephp-codesniffer/CakePHP src/ tests/",
+        "stan": "phpstan analyse src/",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -44,6 +44,8 @@
         ],
         "cs-check": "phpcs --colors -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP src/ tests/",
         "cs-fix": "phpcbf --colors --standard=vendor/cakephp/cakephp-codesniffer/CakePHP src/ tests/",
+        "stan": "phpstan analyse src/",
+        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan:^0.11 && mv composer.backup composer.json",
```
ADmad: It's safer to use https://github.com/phpstan/phpstan-shim instead of phpstan/phpstan.

markstory:
Nice work. We can switch to phpstan-shim next if folks feel it is a good change.

PR #683, opened 2019-05-21
New App Design

cleptric:
A new design for the app template.

Screenshot 2019-05-21 at 20 59 40
Screenshot 2019-05-21 at 20 59 55
Screenshot 2019-05-21 at 21 00 00
Screenshot 2019-05-21 at 21 01 01

The css is based on custom build of milligram, to fix some bugs and adopt the CakePHP CI.
Related PR for bake is cakephp/bake#550.

A missing task is to find a better approach for the related actions.
I don't want to clutter the whole sidebar with actions to related models, so I'm thinking to go into a menu direction for the models only.


ADmad:
Love the updated look and fact that milligram allows for creating clean HTML with minimal extra tags and css classes.

My only concern is milligram is a one man project and doesn't seems to be actively maintained currently. It's last release was over 2 years ago https://github.com/milligram/milligram/releases.

cleptric:
@ADmad That's a valid concern. I checked other mini css frameworks, most of them haven't seen a release for years as well or just need to much work to make look as good as milligram.
The actual amount of sass is very little, so it's pretty easy to change things around.
The obvious bugs I found are already fixed.
I may consider to continue tweaking it and release it as an updated version on my own.

cleptric:
Screenshot 2019-05-21 at 23 01 44
New style for DB text columns

inoas:
All nice and such but what would really improve day-to-day developer experience is classless or just bootstrap-4-grid/responsive utilities classes in cake bake templates and basic app+error layouts instead of the current tag and class soup

cleptric:
@inoas Please have a look at the updated bake templates. The is barely any calls nor tag soup left.
Adding different css should be a task of a few minutes now compared to before.

inoas:
aside the double pagiantor/pagination class this looks really nice now!
"button float-right" seems a little unnecessary. Also I'd go with either always buttons or only for "write" operations

ADmad:
Good to know there's already another framework using milligram. That improves the chance of milligram being maintained in future.

templates/layout/default.php
```diff
@@ -11,7 +11,6 @@
  * @link          https://cakephp.org CakePHP(tm) Project
  * @since         0.10.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
- * @var \App\View\AppView $this
```
dereuromark: Those need to be recovered!

bravo-kernel:
❤️

templates/layout/default.php
cleptric: Will follow up on this

PR #686, opened 2019-06-11
Update error pages

cleptric:
I forgot to add the new style to the debug off error pages.
They look like 

Screenshot 2019-06-11 at 20 22 41

and

Screenshot 2019-06-11 at 20 22 38

### DIFF
--- a/templates/layout/error.php
+++ b/templates/layout/error.php
@@ -23,26 +23,21 @@
     </title>
     <?= $this->Html->meta('icon') ?>
 
-    <?= $this->Html->css('base.css') ?>
-    <?= $this->Html->css('style.css') ?>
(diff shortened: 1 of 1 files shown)


templates/layout/error.php
```diff
@@ -23,26 +23,21 @@
     </title>
     <?= $this->Html->meta('icon') ?>
 
-    <?= $this->Html->css('base.css') ?>
-    <?= $this->Html->css('style.css') ?>
+    <link href="https://fonts.googleapis.com/css?family=Raleway:400,700" rel="stylesheet">
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/normalize.css@8.0.1/normalize.css">
```
saeideng: could we dont use external assets ?

markstory: I think that's a reasonable request.

PR #690, opened 2019-07-25
Use RouteBuilder instance of Router's static methods.

ADmad:


config/routes.php
```diff
@@ -44,61 +46,55 @@
...
  *
  */
-Router::defaultRouteClass(DashedRoute::class);
+$routes->setRouteClass(DashedRoute::class);
 
-Router::scope('/', function (RouteBuilder $routes) {
+$routes->scope('/', function (RouteBuilder $scopedRoutes) {
     // Register scoped middleware for in scopes.
-    $routes->registerMiddleware('csrf', new CsrfProtectionMiddleware([
+    $scopedRoutes->registerMiddleware('csrf', new CsrfProtectionMiddleware([
```
markstory: Could call this `builder` or `scope`.

ADmad: Done

config/routes.php
```diff
@@ -39,66 +41,56 @@
  * Note that `Route` does not do any inflections on URLs which will result in
  * inconsistently cased URLs when used with `:plugin`, `:controller` and
  * `:action` markers.
- *
- * Cache: Routes are cached to improve performance, check the RoutingMiddleware
- * constructor in your `src/Application.php` file to change this behavior.
- *
  */
-Router::defaultRouteClass(DashedRoute::class);
+$routes->setRouteClass(DashedRoute::class);
```
garas: Add phpdoc for `$routes` so autocomplete works

PR #691, opened 2019-07-26
Update development error styling

markstory:
Remove the ugly xdebug stack trace. It was pretty disconnected and kind of ugly.

Companion to cakephp/cakephp#13442


ADmad:
In the past xdebug traces were quite essential for fatal errors. But I guess not so much now that PHP throws exceptions for fatals too.

markstory:
@ADmad Yes. Fatal errors now have normal/useful stack traces. If we find that they don't we can work forwards.

PR #700, opened 2019-08-24
align version check with composer.json definition php: >=7.2

skie:
Fixing template on default app home page.


dereuromark:
Can we DRY this somehow?

ADmad:
I don't think we'll need worry about updating this for at least couple of years :slightly_smiling_face:.

PR #711, opened 2019-09-10
auto-load-app-local

inoas:
fixes #707 (only app_local)

PR #713, opened 2019-09-16
Generate and load app_local.php

inoas:
extended alternative to #711


inoas:
I would also comment out username/password fields in app.php and add a comment behind the line to set these either via the OS environment or app_local.php or .env?

config/app.default.php
```diff
@@ -17,7 +17,7 @@
      * Development Mode:
      * true: Errors and warnings shown.
      */
-    'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),
+    'debug' => filter_var(env('DEBUG', false), FILTER_VALIDATE_BOOLEAN),
```
dereuromark: YES :+1:

config/app.default.php
```diff
@@ -207,15 +207,19 @@
...
-            'username' => null,
-            'password' => null,
+            /*
+             * It is recommended to set these via your environment or app_local.php
+             */
+            //'username' => null,
+            //'password' => null,
             'client' => null,
-            'tls' => null,
+            'tls' => false,
```
inoas: Checked SmtpTransport.php and it is only used a s a boolean.

src/Console/Installer.php
```diff
@@ -181,6 +187,8 @@ public static function setSecuritySalt($dir, $io)
     {
         $newKey = hash('sha256', Security::randomBytes(64));
```
inoas: Should I remove the injection of the salt key into app.php alltogether and only keep it for the generated app_local.php?

config/app.default.php
```diff
@@ -263,9 +267,12 @@
              * the following line and set the port accordingly
              */
             //'port' => 'non_standard_port_number',
-            'username' => 'my_app',
-            'password' => 'secret',
-            'database' => 'my_app',
+            /*
+             * It is recommended to set these via your environment or app_local.php
```
dereuromark: Is it "via" or "through" in English?

config/app_local.default.php
```diff
@@ -0,0 +1,67 @@
+<?php
+/*
+ * Local configuration file to provide any overrides to your app.php configuration.
+ */
+return [
+    /**
```
dereuromark: We shouldnt use docblocks when those are comment blocks.
Docblocks are specificially for variables, methods and alike and require @ tags inside.

That's a misconcept used already in core/app. This can also be done separately though as follow up.

config/app.default.php
ADmad: "via" is also valid but "through" might help non-native English speaker understand better.

markstory: I would go with 'through'.

inoas: i prefer via because it is less ambigious in my mind but I changed it nevertheless.

src/Console/Installer.php
inoas: What about this question. Any core team comments?

inoas:
Anything else?

config/app.default.php
```diff
@@ -207,20 +207,24 @@
...
+             * The keys host, port, timeout, username, password, client and tls
+             * are used in SMTP transports
              */
             'host' => 'localhost',
             'port' => 25,
             'timeout' => 30,
-            'username' => null,
-            'password' => null,
+            /*
+             * It is recommended to set these through your environment or app_local.php
```
garas:
```suggestion
             * It is recommended to set these options through your environment or app_local.php
```

config/app.default.php
```diff
@@ -263,9 +267,12 @@
              * the following line and set the port accordingly
              */
             //'port' => 'non_standard_port_number',
-            'username' => 'my_app',
-            'password' => 'secret',
-            'database' => 'my_app',
+            /*
+             * It is recommended to set these through your environment or app_local.php
```
garas:
```suggestion
             * It is recommended to set these options through your environment or app_local.php
```

src/Console/Installer.php
garas: It should be no secrets in git repo. So `app.php` should have no secrets.

ionas:
Any further requests? Any concerns? Ready to merge?

config/app.default.php
```diff
@@ -92,7 +92,7 @@
         // 'cacheTime' => '+1 year'
```
ionas: Why is this not enabled by default (and in app_local set to +1 second while in app.php set to +1 year?
It is common to long-cache assets via http headers of the web server anyway isnt it?

config/app.default.php
```diff
@@ -238,7 +242,7 @@
         ],
```
ionas: Should Email.default.charset and Email.default.headerCharset not be set to utf-8 by default?

.gitignore
```diff
@@ -1,6 +1,6 @@
 # CakePHP specific files #
 ##########################
-/config/app.php
+/config/app_local.php
```
garas: `app.php` is required by bootstrap, but currently not in git and not in `.gitignore`.

garas: But two editable files is not OK.

It should be `app.default.php`:
- big file with full list of settings,
- secrets/credentials free, 
- tries to use environment,
- always bootstrap loaded,
- git versioned,
- production friendly.

On local machines, it should be only optional file `app.php`:
- small file with minimal set of overrides,
- no environment fallbacks (it is either environment variables or overrides file),
- has secrets and credentials,
- optionally bootstrap loaded,
- git ignored,
- copied from development friendly template `app_local.default.php` (`app.dev.php`/`app.template.php`/...?).

ADmad: Using `app.default.php` + `app.php` is no different from using `app.php` + `app_local.php`. Using the latter combination allows more flexibility for eg. instead of `app_local.php`. I for e.g. use `app_dev.php`,  `app_stagin.php`, `app_prod.php` etc. loaded based on an environment variable. Having the "local" file for each environment named separately give more clarity.

saeideng: if `app.php` is not in `gitignore` and if `secret` should be there 
so in server you should have `app_local.php` ?
if yes why you used `_local` name?

garas: Yes, it could be multiple templates `app_dev.php`, `app_staging.php`, `app_prod.php`.

But untracked copied files should be minimal to be easily reviewable. 

Currently `app.php` is copied from big `app.default.php`, and then it is hard keep big untracked file in sync.

It should be one big git tracked, always loaded file.

Overrides comes through environment and/or small untracked `app.php`/`app_*.php` file.

From current state, it easiest to load `app.default.php + app_local.php`.
But then there is no `app.php`, so I suggest to use `app_local.php` as template for `app.php` and load `app.default.php + app.php`

garas: Personally I  always load only `app.php` (copied from `app.dev.php` template):

```php
<?php
$config = require 'app.default.php';
// add/modify/unset $config keys
return $config;
```

It is easier to understand plain PHP array ops, and there is no  `Hash::merge()`.

ADmad: The template file `app.default.php` should be deleted by the installer after `app.php`. It's useless from that point onwards and can just cause confusion.

ADmad: I prefered it when the template file was named `app.php.default` as it made it clearer that it was a template file and wasn't actually loaded.

garas: But it would restored on next `git checkout`...

ADmad: We can always change the ignore rules as needed.

garas: You can't ignore file if already in git file tree.

dereuromark: @garas Smartest way is two separate files in git. See why at https://sandbox.dereuromark.de/pages/best-practices#config
Then one app_local.php and done. 
Dont make it more complicated than it needs to be for beginning.
Separate files per env people can do themselves.

ADmad: @garas https://www.git-tower.com/learn/git/faq/ignore-tracked-files-in-git

Even if does get checked out for existing apps it doesn't really matter. It's just an extra file which isn't actually loaded.

But at least it will be gone for new app which start from this skeleton.

garas: @ADmad The page from link ignores files by removing files from git tree, because "Git can only ignore files that are untracked".

@dereuromark  So skeleton `app.default.php` should just be renamed to `app.php` here as I see then no point of copy/delete with installer. Like in sandbox. Skeleton would then have `app.php` from start. Then only app_local should be copied from template.

I just used different names:
tracked `app.default.php` + untracked `app.php` like in skeleton
You:
tracked `app.php` + untracked `app_local.php` like in sandbox.

ionas: > `app.php` is required by bootstrap, but currently not in git and not in `.gitignore`.

It is generated from app.default.php
However I can just rename app.default.php to app.php and remove the generation part.
However TBH I would keep the generation part as it is because if someone copies in their app.php as they used to it will still work.

inoas: > `app.php` is required by bootstrap, but currently not in git and not in `.gitignore`.

It is generated from app.default.php
However I can just rename app.default.php to app.php and remove the generation part.
However TBH I would keep the generation part as it is because if someone copies in their app.php as they used to it will still work.

inoas: So did you 3 guys game to a conclusion that I can work on?

saeideng: >Smartest way is two separate files in git. See why at https://sandbox.dereuromark.de/pages/best-practices#config

:+1: 

for `cakephp/app` we can have only 
```
Configure::load('app', 'default', false);
Configure::load('app_custom');
```
`app.php` in git (like `app.php.default` )
`app_custom.php` in gitignore 

`app_custom` is proper name for using in both local and server and less confusing

dereuromark: Noo, the only gitignore file must be named app_local as intuitivly expected.

garas: `app.php` (old `app.default.php`) - git tracked
`app_local.default.php` - git tracked
`app_local.php` - git ignored

inoas: So do we have the conclusion to rename app.default.php to app.php and remove the installer parts that copy that file?

dereuromark: The installer just installs app_local.default now to app_local.

inoas: I have pushed an update, happy?

inoas:
Anything else?

config/app.default.php
dereuromark: Unrelated to this PR.

config/app.default.php
dereuromark: Same.

inoas: ok

config/app_local.default.php
```diff
@@ -0,0 +1,68 @@
+<?php
```
ADmad: I would prefer this template file to not have .php extension, so that it's clear that it's not actually loaded.

dereuromark: Then IDE cant auto hint mistakes when editing it.

ADmad: Why does user need to edit this template file? Isn't this file only used by installer to create `app_local.php`?

dereuromark: Contributors here to app do for example.

ADmad: The template file rarely needs update. There's a vastly greater chance of a newbie thinking this file is actually loaded.

Alternative is we can keep the `.php` extension, add it to .gitignore and make the installer delete it after copying.

dereuromark: Isnt it a quite common standard with .default files?
I dont see the need of overprotection if we have docs here.
So far the confusion was around other topics, never include of the template file.

dereuromark: We can merge this, and have a follow up conversation about the file name ideas there, as those are not related to the issue at hand here IMO.

saeideng: @dereuromark merging into `master` ? or `3.next` ?

dereuromark: Since this is skeleton only relevant we can probably do master here just fine.
Doesnt effect existing apps. Only docs would need to be adjusted accordingly.

htstudios: i am using app_local.example.php so far.

inoas: wordpress uses filename-sample.php
sample or example both work fine. example is a bit longer but has 0 ambiguity where sample has more meanings than example.

config/app_local.example.php
```diff
@@ -1,6 +1,9 @@
 <?php
 /*
  * Local configuration file to provide any overrides to your app.php configuration.
+ * Copy and save this file as app_local.php and make changes as required.
+ * Note: It is not recommended to commit files with credentials such as app_local.php
+ * into source code version control such as git.
```
dereuromark: Git.

config/app_local.default.php
inoas: > Why does user need to edit this template file? Isn't this file only used by installer to create `app_local.php`?

Because you are not foced to use composer create project. You can simply clone and copy the file.

config/app_local.example.php
inoas: Not sure everyone knows. But ok.

src/Console/Installer.php
```diff
@@ -99,7 +99,7 @@ public static function postInstall(Event $event)
     public static function createAppLocalConfig($dir, $io)
     {
         $appLocalConfig = $dir . '/config/app_local.php';
-        $defaultLocalConfig = $dir . '/config/app_local.default.php';
+        $defaultLocalConfig = $dir . '/config/app_local.example.php';
```
dereuromark: If we really also make other side effects here now like renaming .default to .example (which IMO should be a follow up ticket/PR), then we must also adjust all other cases to be in sync here, e.g. https://github.com/cakephp/app/blob/master/config/.env.default

inoas: Done?

inoas:
@ADmad as the example file is now used as a basis for templating we could also just rename it that... template.

.env.template and app_local.template.php

... as it is then a template for both, machine (Installer) and human (so could be the .env file in future where the installer just asks for credentials ans inserts them on template strings as .env)

ADmad:
@ionas Using the word `template` or `example` or `sample` in the file name works for me, as it clarifies the intent of the file and it's less likely that someone will think it's actually loaded/included in the app.

I see you have already used `example` which is good enough for me.

src/Console/Installer.php
```diff
@@ -90,19 +90,19 @@ public static function postInstall(Event $event)
     }
 
     /**
-     * Create the config/app.php file if it does not exist.
+     * Create config/app_local.php files if it do not exist.
```
garas:
```suggestion
     * Create config/app_local.php files if it does not exist.
```

dereuromark: file instead of files then :)

src/Console/Installer.php
garas: Example file is used as template. Variable should use _template_ keyword. Or both variables as _source_ and _destination_.

dereuromark:
We will probably need some docs changes/updates to clarify this.

markstory:
Nice work folks :clap: I can take care of the documentation changes @dereuromark

PR #717, opened 2019-09-26
Add call to Router::fullBaseUrl() in bootstrap.

ADmad:
Refs cakephp/cakephp#13672

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -33,7 +33,6 @@ use Cake\Cache\Cache;
 use Cake\Console\ConsoleErrorHandler;
 use Cake\Core\Configure;
 use Cake\Core\Configure\Engine\PhpConfig;
-use Cake\Core\Plugin;
 use Cake\Database\TypeFactory;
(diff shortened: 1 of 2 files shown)


dereuromark:
Is there a soft migration path coming from 3.x? Like BC fallback etc?
Or do we need migration guide info on top?

markstory:
@dereuromark It looks like this is an optimization to avoid parsing the URL inside of Router. The current `Router::fullBaseUrl()` is still backwards compatible with Configure only style setup.

ADmad:
We should document this is migration guide though for most users upgrading won't have any issue.

The only users that can have a problem with this change are those use use `_ssl` key in URL arrays in commands/shells, since `Router::url()` now directly uses context to get host name instead of calling `Router::fullBaseUrl()` and then parsing the URL.

markstory:
@ADmad What would we put in the migration guide for this? Perhaps to update their bootstrap.php based on what is in the current skeleton?

ADmad:
@markstory Yes

Issue #724, opened 2019-10-28
Update README.md to specify 4x via composer


karmicdice:
This is a (multiple allowed):

* [] bug
* [x ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Application Skeleton Version: 3.8
* Platform and Target: Ubuntu, php 7.4

### What you did
composer create-project cakephp/app --prefer-dist --stability=dev

### What happened
It installs 3.8

### What you expected to happen
The line should read:

`composer create-project cakephp/app:4.x --prefer-dist --stability=dev`

PR #725, opened 2019-10-28
Update README.md on 4x tree

karmicdice:
Update README.md on 4x tree to reflect 4x installation via Composer.

Refers to #724

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,12 @@ In case you want to use a custom app dir name (e.g. `/myapp/`):
 composer create-project --prefer-dist cakephp/app myapp
 ```
 
+If you want to install 4.x branch in your custom app directory
+
+```bash
+composer create-project --prefer-dist --stability=dev cakephp/app:4.x myapp
+```
+
 You can now either use your machine's webserver to view the default home page, or start
 up the built-in webserver with:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,6 +25,12 @@ In case you want to use a custom app dir name (e.g. `/myapp/`):
 composer create-project --prefer-dist cakephp/app myapp
 ```
 
+If you want to install 4.x branch in your custom app directory
+
+```bash
+composer create-project --prefer-dist --stability=dev cakephp/app:4.x myapp
```
dereuromark: Not sure if we should duplicate, instead better modify the existing row maybe?
What do others think?

karmicdice: You're right. This isn't master branch. I am directly requesting this pull on 4x branch. It is implied that the user is looking at 4x branch and wants to use 4x. I will make the change.

saeideng: I'm not sure you need to `--stability=dev` 

it works without this , let `cakephp/app` handles this

jdecode:
Why would `composer create-project --prefer-dist cakephp/app:4.x` be still failing?

Getting `Could not find package cakephp/app with version 4.x` error.

Screenshot from 2019-11-07 23-44-17


Adding the flag `--stability=dev` gets the job done though.

Thanks

karmicdice:
> Adding the flag --stability=dev gets the job done though.

Because 4.x is at dev?

jdecode:
> > Adding the flag --stability=dev gets the job done though.
> 
> Because 4.x is at dev?

Shouldn't the documentation reflect what is available (and should work without having to check if the branch is in dev or GA stage?)

karmicdice:
Hmm. Ok.. See... This is currently pushed to 4.x branch which is dev. When user explicitly comes on github and changes the branch to 4.x- this is the readme they will see. It is implied to the user that what he is looking at is dev. The README at **master** will require telling user not when user knows he is at a different branch already.

PR #727, opened 2019-11-05
Replacing overly assuming verbiage in `.env` section of `bootstrap.php` with informational resource

groovenectar:
Replacing overly assuming or opinionated verbiage in the comments by providing an informational link to the `php-dotenv` library repository, which links to https://12factor.net/config

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -48,9 +48,7 @@ use Cake\Utility\Security;
  * You should copy `config/.env.example to `config/.env` and set/modify the
  * variables as required.
  *
- * It is HIGHLY discouraged to use a .env file in production, due to security risks
- * and decreased performance on each request. The purpose of the .env file is to emulate
- * the presence of the environment variables like they would be present in production.
+ * More information: https://github.com/josegonzalez/php-dotenv
  */
 // if (!env('APP_NAME') && file_exists(CONFIG . '.env')) {
 //     $dotenv = new \josegonzalez\Dotenv\Loader([CONFIG . '.env']);


othercorey:
How is discouraging use "overly assuming or opinionated"?

groovenectar:
> How is discouraging use "overly assuming or opinionated"?

I don't understand what you mean... You're saying that my wording in the PR is discouraging?

othercorey:
> I don't understand what you mean... You're saying that my wording in the PR is discouraging?

You said:

>Replacing overly assuming or opinionated verbiage in the comments

I'm asking how discouraging use of .env file in production is "overly assuming or opinionated".

groovenectar:
> > I don't understand what you mean... You're saying that my wording in the PR is discouraging?
> 
> You said:
> 
> > Replacing overly assuming or opinionated verbiage in the comments
> 
> I'm asking how discouraging use of .env file in production is "overly assuming or opinionated".

If someone is uncommenting the lines in order to use an environment file pattern, they must be aware of at least the basics of using `.env` files... Since it's not enabled by default, they're specifically looking to make use of that pattern...

Or if they're tinkering, the PR directs them to a link for the parser library with more information.

It is technically no more insecure to use a `.env` file in production than using `app.php` for configuration, since they are both outside the `webroot`. It will be a bit less performant since it is reading the file from the filesystem once on every request, which can still be a pretty fast process.

At any rate, "HIGHLY DISCOURAGED" is a bit strong, yes?

othercorey:
> If someone is uncommenting the lines in order to use an environment file pattern, they must be aware of at least the basics of using `.env` files... Since it's not enabled by default, they're specifically looking to make use of that pattern...

It seems like this is assuming even more.  Why remove the recommended practice documentation?

If the user is willing to use .env files in production, they can ignore the comment.  If they are unaware of the suggestion, they are making an uninformed decision.
 
> At any rate, "HIGHLY DISCOURAGED" is a bit strong, yes?

This PR removes all comments on the topic though.  There's no benefit to removing something that directs them to investigate their use case.

This is similar to using a secret key in a file.  You typically do not want such information in a repo.  You want to attach as part of the deploy process or configure to look up via a secure api.

Users need to start with the basic suggestion and then make an informed decision about their production environment and deploy process.

There are many users who work with CakePHP without extensive experience in devops and production environments.

groovenectar:
> > If someone is uncommenting the lines in order to use an environment file pattern, they must be aware of at least the basics of using `.env` files... Since it's not enabled by default, they're specifically looking to make use of that pattern...
> 
> It seems like this is assuming even more. Why remove the recommended practice documentation?

It is based on this: https://12factor.net/config
Where on that page does it mention that the removed part of the comment is recommended practice?

> If the user is willing to use .env files in production, they can ignore the comment. If they are unaware of the suggestion, they are making an uninformed decision.
> 
> > At any rate, "HIGHLY DISCOURAGED" is a bit strong, yes?
> 
> This PR removes all comments on the topic though. There's no benefit to removing something that directs them to investigate their use case.

There was no link before, wouldn't a link to documentation allow the user to make an even more informed decision?

The comment contained incorrect information. Using a `.env` file in production does not likely introduce a greater security risk. The variables remain in the context of the PHP process, unlike some setups using system variables in production. There is no way that all security practices can or should or need to be covered in this comment.

> This is similar to using a secret key in a file. You typically do not want such information in a repo. You want to attach as part of the deploy process or configure to look up via a secure api.
>
> Users need to start with the basic suggestion and then make an informed decision about their production environment and deploy process.
> 
> There are many users who work with CakePHP without extensive experience in devops and production environments.

It's standard practice to not commit the `.env` files.

Are you suggesting that comment should contain additional information, guiding users on accepted practices for managing repositories?

groovenectar:
The comment already contains this:

    Uncomment block of code below if you want to use `.env` file during development

Suggesting that it is for use during development

othercorey:
Seems like I edited the last comment instead of quoting it. I changed it back to the original.  My actual reply below.

>Are you suggesting that comment should contain additional information, guiding users on accepted practices for managing repositories?

No.  It's just a very simple warning to users that are looking at the app skeleton as a starting point.  If a new developer was going to copy their local dev setup to production, this helps them evaluate that workflow.

If you are using .env files in production, there is no limitation placed on anything by this.

ADmad:
@groovenectar CakePHP is an opinionated framework. So that comment is stating our opinion about .env files in production. Users are still free to do as they please.

Thank you for your interest but there's no need to change that comment.

groovenectar:
I have experienced a case where that exact comment caused a problem.

And you can't say a case where it has helped, because it does not help.

But fair enough about opinionated, and it's just a comment about code that ships commented.

dakota:
I'm curious how a comment discouraging use in production could cause a problem?

groovenectar:
> I'm curious how a comment discouraging use in production could cause a problem?

If an application does get to production with this pattern, then there is no security issue and the application will run without a problem. There are popular applications that can ship this way. The application will also be in a state where it can scale very nicely in a number of directions including plenty of directions that would involve removing the file in deployment perhaps shaving some nanoseconds.

The comment is not simply discouraging, it is over-the-top with all caps. It is also misleading. If the comment is noticed after-the-fact, and it is taken on face value without consideration about its legitimacy, then it can cause bigger problems via reaction.

So while it can make sense to use it only for development, and it is my opinion as well that it is ideal to use it that way, the comment as-worded can cause problems in some situations where it does get deployed. A reaction resulting from that specific comment, because it is not worded as simply discouraging, may cause regression on entire systems.

othercorey:
@gr

> I have experienced a case where that exact comment caused a notable problem.
> 
> And you can't say a case where it has helped, because it does not help.

We value feedback.  However, the PR you submitted simply removed the comment without any background information or example.  It was clear you preferred that method and so wanted the framework to support that.

We cannot replace one assumption with other.  Either the user is completely unaware of their decisions or they are fully aware and know how to implement those choices.

If you want to augment the explanation in the comment with something useful (and provide the link you had before), that might be a better discussion.  A large paragraph on the topic will not work though.

jchook:
Seriously though, the "config in the environment" part of 12factor doesn't ever give any guidelines on how to accomplish it securely, or why it provides extra security.

Several articles I have found express concerns about the [security](https://medium.com/poka-techblog/the-best-way-to-store-secrets-in-your-app-is-not-to-store-secrets-in-your-app-308a6807d3ed) of storing secrets in the environment.

othercorey:
@groovenectar I updated the comment in 2.x branch to give more info.  The best recommendation I can give any team is to communicate and discuss deploy strategy.  I can't help if developers give different best practices or recommendations as there are many ways deploy/configure apps.

groovenectar:
@othercorey Your update in the 2.x branch looks great. I can see where my comments made it seem like I preferred something different than the general thoughts on what may be ideal. But more coming from an angle of preventing unwarranted overreaction. If your wording had been in there, I'm sure that things would have gone much smoother and a huge regression would have been avoided. Thank you for taking the time to update it.

PR #731, opened 2019-11-15
Add `--stability=dev` flag to allow `dev` support

jdecode:
By default the composer will try to install the `stable` 4.x version, and in absence of which the error is triggered specifying that the package `cakephp/app` with version 4.x is not available/found.
This happens while 4.x is `dev`, not stable.

Add `--stability=dev` flag to allow `dev` version to be installed by default, until the stable version is released.


markstory:
Thank you 👍

Issue #730, opened 2019-11-14
Creating projects creates two sets of placeholder default credentials


othercorey:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Application Skeleton Version: 4.x

### What you did
Run project setup from tutorial.

```
php composer.phar create-project --prefer-dist cakephp/app bookmarker
```

### What happened
app_local.example.php is copied to app_local.php

Both app_local.php and app.php have full placeholder db credentials.

### What you expected to happen
Just app_local.php to have db credentials as users are pushed to use that file.

othercorey:
@ADmad as discussed, if you want something to track.

ADmad:
Besides db credentials there are various other config like for email which too are duplicated. This can lead to confusion in future as someone might make changes in `app.php` not realizing the defaults being overwritten in `app_local.php`.

There's should be no / minimal overlap of defaults set in the 2 files with config which generally vary depending upon environment being in `app_local.sample.php` and others in `app.php`.

markstory:
Could we have only credentials and hosts/ports defined in app_local?

othercorey:
If we move the defaults, I think app.php should contain a comment in each section just stating that the settings are in app_local.php and to use app.php for deploy.

You don't want just a blank app.php as that doesn't help a new user.

ADmad:
> and to use app.php for deploy

No, `app.php` is commit to the repo so it should not have any credentials. The server should have it's own `app_local.php`.

othercorey:
> > and to use app.php for deploy
> 
> No, `app.php` is commited to the repo so it should not have any credentials. The server should have it's own `app_local.php`.

Sorry, I meant deployed with ENV configuration that app

dereuromark:
This is now resolved?

PR #732, opened 2019-11-16
4.x - Removed duplicate datasource configurations from app.php

othercorey:
https://github.com/cakephp/app/issues/730

I removed duplicate datasource settings from app.php and app_local.php.  I added a few more comments and spaced things out.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -259,32 +259,27 @@ return [
      *   other RDBMS.
      */
     'Datasources' => [
+        /**
+         * These configurations should contain permanent settings used
+         * by all environments.
+         *
+         * The values in app_local.php will override any values set here
+         * and should be used for local and per-environment configurations.
+         *
+         * Environment variable based configurations can be loaded here or
+         * in app_local.php depending on the applications needs.
+         */
         'default' => [
             'className' => Connection::class,
             'driver' => Mysql::class,
             'persistent' => false,
-            'host' => 'localhost',
-            /**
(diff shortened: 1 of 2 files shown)


config/app.php
```diff
@@ -312,8 +307,6 @@
              * which is the recommended value in production environments
              */
             //'init' => ['SET GLOBAL innodb_stats_on_metadata = 0'],
-
-            'url' => env('DATABASE_URL', null),
```
garas: `env('DATABASE_URL', null)` should leave in `app.php`, so it would work with environment variables, but no `app_local.php`.

othercorey: This point of this change is to not have hidden db credentials in app.php.  Users should add 'url' if they want DATABASE_URL to permanently override.

dereuromark: Having it stubbed in the local file sounds good.

jdecode: While I believe (actually, I hope) that I understand the reason for having app_local.php, but wouldn't it make more sense to push (and promote) using .env file for local development?

Understanding that this might mean that more developers be educated about using environment variables in non-familiar hosting setups, but certain clear ideas can definitely be floated how to use .env in local setups while using same code base (no ignoring any other files, apart from .env file) and have a same-code setup on all environments (Local, Dev, QA, UAT, Prod).

ADmad: Not everyone can / want to use environment variables in production. For e.g. apps deployed on shared hosting.

PR #735, opened 2019-11-20
Add missing throws tag

davidyell:
As the internals of loadComponent can throw an exception it should be documented, so a use can choose to catch it

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -36,6 +36,7 @@ class AppController extends Controller
      * e.g. `$this->loadComponent('Security');`
      *
      * @return void
+     * @throws \Exception If the component cannot be found
      */
     public function initialize(): void
     {


src/Controller/AppController.php
```diff
@@ -36,6 +36,7 @@ class AppController extends Controller
      * e.g. `$this->loadComponent('Security');`
      *
      * @return void
+     * @throws \Exception If the component cannot be found
```
dereuromark: Pretty sure it is not a root exception, but Cake specific one.
Also, these things are hard to keep in sync if we allow those outside of actual 0-level usage (inside the same method).
So not a big fan of this. 

And finally, those kind of exceptions you definitely should never document here, those are internal 500 developer things that must never happen under normal circumstances. If anything, you document useful ones that you are allowed to watch out and even catch if necessary inside user land code.

davidyell: It was suggested by PHP Storm, so I am not overly attached to having it be merged. If the preference is to omit it, that's okay.

dereuromark: For these kind of, for sure.
The IDE can also be configured to level 0, which I would recommend for these things.

ADmad:
Closing for reason @dereuromark stated.

PR #738, opened 2019-11-26
4.x - move ConsoleShell to under command namespace

saeideng:


ADmad:
Do people use this REPL shell? I have never heard anyone mention using it. The lib the Console shell relies on pulls in a symfony dependency which overrides our debug() function.

dakota:
I personally use it multiple times a day. Would be sad to see it go.

Also, I prefer the symfony `debug()` and `dd()` output (it's super interactive and easier to find what I'm looking for)

dereuromark:
@dakota Wouldnt a plugin solution suffice here?

dereuromark:
> Also, I prefer the symfony debug() and dd() output (it's super interactive and easier to find what I'm looking for)

I agree that they have the better debug tooling here in terms of usability, but secretly overwriting the core ones is a nogo IMO. This should be a clear opt-in behavior you want to have, rather than getting this by accident right now.

ADmad:
> I agree that they have the better debug tooling here in terms of usability, but secretly overwriting the core ones is a nogo IMO. This should be a clear opt-in behavior you want to have, rather than getting this by accident right now.

Exactly.

markstory:
I also use the repl shell quite a bit.

saeideng:
we can remove the class and docs if necessary later 
I merged for now , we have to mention in docs before removing ( if needed )

PR #744, opened 2019-12-04
Change to require cakephp/migrations

nojimage:
Move to `require` cakephp/migrations. Because it need on CLI.

https://github.com/cakephp/app/blob/4.x/src/Application.php#L98


rchavik:
Why? And shouldnt this be on require-dev?

dereuromark:
No it should be installable on prod.

PR #763, opened 2019-12-28
Update to phpstan v0.12

ADmad:


ADmad:
Travis config was broken and jobs for phpcs and phpstan weren't running properly. Instead of fixing it I replaced it with GA workflow.

PR #768, opened 2019-12-30
Update bake's dependency constraint.

ADmad:
This ensures stable version of TwigView is used.


ADmad:
I also updated constraints for other dependencies.

PR #776, opened 2020-01-26
Updating Deprecated setMethods()

challgren:
Updating Deprecated `setMethods` when mocking Application for testing. 

See https://github.com/sebastianbergmann/phpunit/pull/3687


markstory:
Do we also need to update the minimum require phpunit version?

challgren:
@markstory I don't think so since cakephp/app provides `^8.5` and onlyMethods is available in 8.3, 8.4 and 8.5

https://github.com/sebastianbergmann/phpunit/blob/8.5/src/Framework/MockObject/MockBuilder.php#L210
https://github.com/sebastianbergmann/phpunit/blob/8.4.0/src/Framework/MockObject/MockBuilder.php#L213
https://github.com/sebastianbergmann/phpunit/blob/8.3.0/src/Framework/MockObject/MockBuilder.php#L208

PR #779, opened 2020-02-12
Add Debugger configure block.

markstory:
Related to cakephp/docs#6462


config/app.php
```diff
@@ -184,6 +184,22 @@
...
+     *
+     * - `editor` Choose the which editor URL format you want to use.
+     *   By default atom, emacs, macvim, phpstorm, sublime, textmate, and vscode are
+     *   available. You can add additional editor link formats using
+     *   `Debugger::addEditor()` during your application bootstrap.
+     * - `outputMask` A mapping of `key` to `replacement` values that
+     *   `Debugger` should replace in dumped data and logs generated by `Debugger`.
+     */
+    'Debugger' => [
+        'editor' => 'phpstorm',
```
ADmad: I wonder if we can get some sponsorship from phpstorm for having their editor as default? :slightly_smiling_face:

config/app.php
```diff
@@ -184,6 +184,22 @@
         'trace' => true,
     ],
 
+    /**
+     * Debugger configuration
+     *
+     * Define development error values for Cake\Error\Debugger
+     *
+     * - `editor` Choose the which editor URL format you want to use.
```
othercorey:
```suggestion
     * - `editor` Set the editor URL format you want to use.
```

config/app.php
markstory: I assume this is a default value that will be useful to more people than my personal preferences :smile:

ADmad: I would have preferred any other open source editor/ide, for e.g. vscode, but meh.

PR #780, opened 2020-02-16
Override Milligram fonts and anchor elements

mikebirch:
Override the fonts and weights set by Milligram. Use colours for anchor elements that are more accessible. Match the styles with other CakePHP sites like https://book.cakephp.org


markstory:
Thanks!

PR #785, opened 2020-04-18
Allow unstable releases only for migrations plugin

ADmad:


dereuromark:
Nifty trick :)
once we release stable, does this still use beta then? or does it auto jump to the stable?

othercorey:
It should jump to the stable unless something else requires a beta release.

dereuromark:
Now we need to make sure the eco system is ready, and all stable releases out there :)

PR #786, opened 2020-04-22
[4.next] REPL Convenience Functions

ravage84:
This is the result of the discussions in https://github.com/cakephp/cakephp/pull/14490

This in combination with https://github.com/cakephp/cakephp/pull/14491 is an alternative to https://github.com/cakephp/cakephp/pull/14490.

We might want to think about extracting the REPL to its own plugin. We would move the files from this PR with it.

### DIFF
--- a/src/Command/ConsoleCommand.php
+++ b/src/Command/ConsoleCommand.php
@@ -51,6 +51,8 @@ class ConsoleCommand extends Command
             return static::CODE_ERROR;
         }
 
+        include 'repl_functions.php';
+
         $io->out("You can exit with <info>`CTRL-C`</info> or <info>`exit`</info>");
         $io->out('');
 
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,10 @@
         "psr-4": {
             "App\\Test\\": "tests/",
             "Cake\\Test\\": "vendor/cakephp/cakephp/tests/"
-        }
+        },
+        "files": [
+            "src/Command/repl_functions.php"
+        ]
     },
     "scripts": {
         "post-install-cmd": "App\\Console\\Installer::postInstall",
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,6 +1,8 @@
 parameters:
     level: 7
     checkMissingIterableValueType: false
+    autoload_files:
+        - config/paths.php
     excludes_analyse:
         - src/Console/Installer.php
     ignoreErrors:
--- a/src/Command/repl_functions.php
+++ b/src/Command/repl_functions.php
@@ -1,0 +1,63 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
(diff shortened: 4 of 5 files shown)


dereuromark:
The plugin has the advantange of auto pulling fixes/updates
Right now this would require manual merge from this project code.

src/Command/ConsoleCommand.php
```diff
@@ -51,6 +51,8 @@ public function execute(Arguments $args, ConsoleIo $io)
             return static::CODE_ERROR;
         }
 
+        include APP . 'repl_functions.php';
```
ravage84, line 54: PHPStan says: :man_shrugging: 

> Constant APP not found.   

https://github.com/cakephp/app/runs/609209541

dereuromark: instead of root I would probably put it into /src/Console/

ADmad: Some projects use `src/functions` folder to hold functions files. I kinda like that.

ADmad: > Constant APP not found.

Use `autoload_files` config of phpstan to load `tests/bootstrap.php` like here https://github.com/cakephp/cakephp/blob/c68b60be0198ecc534ef84d08f84007668ad4808/phpstan.neon#L9
You can also autoload `repl_functions.php` functions that way about modifying `composer.json`.

ravage84:
> The plugin has the advantage of auto pulling fixes/updates
> Right now this would require manual merge from this project code.

@dereuromark what would be the process to extract it?

markstory:
Creating a repl plugin would be a good strategy. The console command has been tedious to maintain and do upgrades on. Having a new plugin would also allow us to add more convenience wrapper functions that aren't needed in most production deploys.

ravage84:
@markstory sounds good. What is the process of such a code extraction into a plugin? I'm used to do such, but not when it comes to the core code. I mean in hinsight of the sub-split, e.g. https://github.com/cakephp/cakephp/pull/6565/files? If I get some guidance, I can try to take stab at it.

markstory:
> What is the process of such a code extraction into a plugin? I'm used to do such, but not when it comes to the core code. 

We don't really have a well defined process as we've not done this specific change before.

I would create the plugin first with all the desired functions and the ConsoleCommand from the application skeleton. Then we can mark all the moved functions in cakephp/cakephp and cakephp/app as deprecated, and suggest the new plugin.

dereuromark:
@markstory You have the necessary rights to create this new plugin, we dont.

markstory:
I've made a https://github.com/cakephp/repl repository and given the core team access to it :smile:

othercorey:
Do we want to make this move for 4.1?

markstory:
@othercorey I would prefer we flush out the repl plugin and add it as a dependency to the application. So far we have a few things that can go in that plugin:

* The current repl command including the psy/shel dependency.
* New function wrappers that are useful in a repl scenario.

ADmad:
Closing for reason stated above.

dereuromark:
@ravage84 Is there any progress on the plugin?
Would also solve https://github.com/cakephp/docs/pull/6657/files#diff-e560fba069f2363505992d003b798b24R20 as a more convenient call to get tables.

ravage84:
@dereuromark I'll try to take a stab at it this week.

dereuromark:
@ravage84 You still wanna make the plugin? https://github.com/cakephp/repl is still empty and ready to get filled with useful repl tooling.

ravage84:
@dereuromark yeah, I'm still interested. Unfortunately, I'm swamped with work and other stuff, lately.
I still have it on my TODO list.

ADmad:
https://github.com/cakephp/repl/commit/fbd6843771fdd9be6cbf93e7380511c35505a196

ravage84:
@ADmad thanks for taking care.

markstory:
Thanks @ADmad 👏 :clap:

PR #788, opened 2020-05-09
Remove unreachable statement

o0h:
This is a very small fix.
I found a line that couldn't be executed in `PagesController::display()`, so cleaned up it.

NOTE:
In [this commit](https://github.com/cakephp/app/commit/0bc840e92dd4c2a18ffcc8b354fa663964367c5d), this action has been updated to return in early.


othercorey:
Thanks!

Issue #771, opened 2020-01-14
Suggested improvements for refreshed design 


mikebirch:
* CakePHP Application Skeleton Version: 4.0.2

The refreshed design is lighter and looks good, but I have noticed some issues:

- the color of anchor elements (including hover and focus styles) means that they're hard to distinguish from other text. Maybe use red and blue on hover like https://cakephp.org/ (for body text, not actions and menus).
- Raleway font is used for anchor elements, but Roboto for everything else. A lightweight design could look good by using one or the other, but doesn't need both.
- There is no spacing between labels and radio buttons

cleptric:
@mikebirch Thanks for the feedback, I'll try to make some improvements here.

ADmad:
@cleptric Can this be closed?

cleptric:
IMO, yes.
Using different colours for the anchor tags might be a question of personal preference.
Removing the font is easy for a person to do.
I like the slight variation, so I would keep it as is.

mikebirch:
The colours of anchor elements is an accessibility issue, not just personal preference, but I guess it’s easy enough to override.

markstory:
@mikebirch Any chance you'd be up to open a pull request to pick a better link color?

ndm2:
While we're at it, how about interchanging button colors, or at least using a different highlight color? Currently buttons are red when they're inactive, and grey when they're active, I find that rather counter intuitive.

mikebirch:
> @mikebirch Any chance you'd be up to open a pull request to pick a better link color?

Yes, have had a go at it and tidied up the typography too. Raleway (400 and 700 weight) is included in src/templates/layout/default.php but Roboto is not. Headings are now set in Raleway and all text is 400 weight. Anchor elements are blue except for navigation, table headings and action links.

PR #792, opened 2020-05-14
Use a single font family.

ADmad:
This keeps things simple and avoids pulling in additional external resource.
Refs #771.

### DIFF
--- a/templates/layout/error.php
+++ b/templates/layout/error.php
@@ -23,7 +23,6 @@
     </title>
     <?= $this->Html->meta('icon') ?>
 
-    <link href="https://fonts.googleapis.com/css?family=Raleway:400,700" rel="stylesheet">
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/normalize.css@8.0.1/normalize.css">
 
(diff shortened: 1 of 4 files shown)


markstory:
Using raleway was nice as it pulls part of the CakePHP branding elements from the website/docs into the application.

templates/layout/error.php
```diff
@@ -23,7 +23,6 @@
     </title>
     <?= $this->Html->meta('icon') ?>
 
-    <link href="https://fonts.googleapis.com/css?family=Raleway:400,700" rel="stylesheet">
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/normalize.css@8.0.1/normalize.css">
```
dereuromark, line 26: This one is still external right?
So maybe we don't have to remove the railway - if we cant get external resources away anyway.

ADmad: I am not sure why `normalize.css` is used through CDN. It too should be added to webroot/css itself.

ADmad:
I don't have a strong opinion on this either way, hence closing.

PR #800, opened 2020-07-05
Restrict updates to bugfixes releases only by default.

ADmad:
New minors for CakePHP have caused issues in the past inspite of the API being backwards compatible.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,17 +6,17 @@
     "license": "MIT",
     "require": {
         "php": ">=7.2",
-        "cakephp/cakephp": "~4.1",
+        "cakephp/cakephp": "~4.1.0",
         "cakephp/migrations": "^3.0",
         "cakephp/plugin-installer": "^1.2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -6,17 +6,17 @@
...
         "mobiledetect/mobiledetectlib": "^2.8"
     },
     "require-dev": {
         "cakephp/bake": "^2.0.3",
-        "cakephp/cakephp-codesniffer": "~4.2",
+        "cakephp/cakephp-codesniffer": "~4.2.0",
         "cakephp/debug_kit": "^4.1",
         "josegonzalez/dotenv": "^3.2",
-        "phpunit/phpunit": "^8.5",
+        "phpunit/phpunit": "~8.5.0",
```
dereuromark, line 19: There was no need to limit those require-dev IMO.

ADmad: I synced it with core.

PR #804, opened 2020-08-26
Fix api url in default layout

blieb:
At the moment the api url in the default layout goes to a 404 page. Changed the link to a working url.

### DIFF
--- a/templates/layout/default.php
+++ b/templates/layout/default.php
@@ -42,7 +42,7 @@ $cakeDescription = 'CakePHP: the rapid development php framework';
         </div>
         <div class="top-nav-links">
             <a target="_blank" rel="noopener" href="https://book.cakephp.org/4/">Documentation</a>
-            <a target="_blank" rel="noopener" href="https://api.cakephp.org/4/">API</a>
+            <a target="_blank" rel="noopener" href="https://api.cakephp.org/4.1/">API</a>
         </div>
     </nav>
     <main class="main">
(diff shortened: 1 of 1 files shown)


templates/layout/default.php
```diff
@@ -42,7 +42,7 @@
         </div>
         <div class="top-nav-links">
             <a target="_blank" rel="noopener" href="https://book.cakephp.org/4/">Documentation</a>
-            <a target="_blank" rel="noopener" href="https://api.cakephp.org/4/">API</a>
+            <a target="_blank" rel="noopener" href="https://api.cakephp.org/4.1/">API</a>
```
dereuromark, line 45: Additionally, the existing URL should probably 301 to the latest.

ADmad, line 45:
```suggestion
            <a target="_blank" rel="noopener" href="https://api.cakephp.org/">API</a>
```

Homepage will direct to latest minor.

blieb: Ah yes that is a better solution then :)

blieb: Hmm yes. that is maybe also what I expect. I see this also don't work for /3/

templates/layout/default.php
```diff
@@ -42,7 +42,7 @@
         </div>
         <div class="top-nav-links">
             <a target="_blank" rel="noopener" href="https://book.cakephp.org/4/">Documentation</a>
-            <a target="_blank" rel="noopener" href="https://api.cakephp.org/4/">API</a>
+            <a target="_blank" rel="noopener" href="https://api.cakephp.org/">API</a>
```
markstory, line 45: The `/4.latest/` might be a better path that redirects correctly too.

dereuromark:
I guess for now this seems already better and working.

PR #809, opened 2020-10-01
Update README.md

arjit-gupta:


ADmad:
@arjit-gupta Stop being a part of this problem https://blog.domenic.me/hacktoberfest/.

PR #811, opened 2020-10-09
Centralize configuration for cache,log and database driver settings

rchavik:
It should be easier to change these configuration to the top of file so that we don't have to fish out the values below.
Having the $cachePrefix as variable also avoids forgotting to change a prefix in one of the config.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -1,14 +1,17 @@
 <?php
 
-use Cake\Cache\Engine\FileEngine as CacheEngine;
+use Cake\Cache\Engine\FileEngine;
 use Cake\Database\Connection;
-use Cake\Database\Driver\Mysql as DatabaseDriver;
+use Cake\Database\Driver\Mysql;
 use Cake\Error\ExceptionRenderer;
-use Cake\Log\Engine\FileLog as LogEngine;
+use Cake\Log\Engine\FileLog;
 use Cake\Mailer\Transport\MailTransport;
 
 $appName = 'myapp';
 $cachePrefix = "${appName}_";
+$logEngine = FileLog::class;
+$cacheEngine = FileEngine::class;
+$databaseDriver = Mysql::class;
 
 return [
     /*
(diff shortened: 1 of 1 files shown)


ADmad:
I don't like the unnecessary aliasing of classes.

rchavik:
Any other way of doing this without aliasing?

ADmad:
You can use env vars or URL based config instead. 

In general one can modify the config file in whatever way they need but let's keep the default one simple so that it's easy to grok for new users.

rchavik:
> You can use env vars or URL based config instead.

I would argue that these are actually advanced features.

> In general one can modify the config file in whatever way they need but let's keep the default one simple so that it's easy to grok for new users.

I think changing one line instead of going through the file to change all 4 occurences of `FileEngine` to `RedisEngine` is simpler, or 3 lines Log Engine.

config/app.php
```diff
@@ -1,12 +1,15 @@
...
 use Cake\Database\Connection;
-use Cake\Database\Driver\Mysql;
+use Cake\Database\Driver\Mysql as DatabaseDriver;
 use Cake\Error\ExceptionRenderer;
-use Cake\Log\Engine\FileLog;
+use Cake\Log\Engine\FileLog as LogEngine;
 use Cake\Mailer\Transport\MailTransport;
 
+$appName = 'myapp';
+$cachePrefix = "${appName}_";
```
markstory, line 11: I like this consolidation, I'm less sure about the class aliasing being simpler for folks new to CakePHP.

rchavik:
Adjusted to not use class aliasing.

rchavik:
are we going to drop this?

markstory:
I like the prefix consolidation. I'm not sure that the classnames need the same treatment, but I don't think I'm the intended audience for this change.

ADmad:
I don't like using variables for classname either.

These changes are more a personal preference then a general improvement and I don't think we need to change anything.

rchavik:
Closing then. Thanks.

PR #815, opened 2020-11-04
Update .gitattributes

markusramsak:
several file types added

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,6 +11,7 @@
 *.sql text
 *.md text
 *.po text
+*.pot text
 *.js text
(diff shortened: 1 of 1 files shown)


dereuromark:
Done in https://github.com/cakephp/app/commit/2cd592ee1f4ba8e8f0a8f0aa4751c8c69577a6cc
There is no need to define the defaults
As long as nothing else is declared those are treated as text.

markusramsak:
what about those added binary file types?

dereuromark:
Sry, you are right.
You can make a new PR that we can merge.

PR #822, opened 2020-11-19
Use string for time type

dereuromark:
See current bugs and hacks needed as per https://github.com/cakephp/chronos/issues/102#issuecomment-726750246
and also discussion of https://github.com/cakephp/cakephp/issues/13440

It seems until a good value object is found, a string default seems like the right choice for now.

If so, should this go into 4.next?

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -34,6 +34,8 @@ require CORE_PATH . 'config' . DS . 'bootstrap.php';
 use Cake\Cache\Cache;
 use Cake\Core\Configure;
 use Cake\Core\Configure\Engine\PhpConfig;
+use Cake\Database\TypeFactory;
+use Cake\Database\Type\StringType;
 use Cake\Datasource\ConnectionManager;
 use Cake\Error\ConsoleErrorHandler;
 use Cake\Error\ErrorHandler;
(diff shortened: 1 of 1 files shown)


markstory:
I think this could be ok in master. The current mapping for time columns aren't great.

config/bootstrap.php
```diff
@@ -187,22 +187,23 @@
...
-// TypeFactory::build('timestampfractional')
+// \Cake\Database\TypeFactory::build('timestampfractional')
 //    ->useMutable();
-// TypeFactory::build('datetimetimezone')
+// \Cake\Database\TypeFactory::build('datetimetimezone')
 //    ->useMutable();
-// TypeFactory::build('timestamptimezone')
+// \Cake\Database\TypeFactory::build('timestamptimezone')
 //    ->useMutable();
+\Cake\Database\TypeFactory::map('time', \Cake\Database\Type\StringType::class);
```
ADmad, line 206: If you add a `use` statement for the class you won't have to update the other examples above either.

othercorey:
Can we update and merge this?

dereuromark:
I appreciate any help here ATM.

PR #823, opened 2020-11-27
Update .gitignore

markusramsak:
typo?!


garas:
No.

Windows XP Media Center Edition creates `ehthumbs.db` which holds previews of video files [wikipedia](https://en.wikipedia.org/wiki/Windows_thumbnail_cache).  It is old OS, I don't know if any newer OS still uses it.

PR #829, opened 2021-03-08
Use repl package and remove mobiledetect

markstory:
Instead of shipping the console command with the application skeleton using a separate package will allow us to deliver fixes to the console command in a reasonable way. It also enables us to provide more shortcut functions for use in the repl and development mode.

I've removed mobiledetect as a direct dependency. I don't think it gets a ton of usage by users and it can live on as a comment and optional package that can be installed as necessary.


src/Application.php
```diff
@@ -128,8 +128,10 @@ public function services(ContainerInterface $container): void
     protected function bootstrapCli(): void
     {
         try {
+            $this->addPlugin('Cake/Repl');
```
othercorey, line 131: Shouldn't this try one plugin at a time?

dereuromark: In this case, yes.
The grouping only makes sense if the following can only exist or be available with the first.

src/Application.php
```diff
@@ -128,8 +128,10 @@ public function services(ContainerInterface $container): void
...
             $this->addPlugin('Bake');
         } catch (MissingPluginException $e) {
+            debug($e->getMessage());
```
ADmad, line 134: Including `config/bootstrap` file as done [here](https://github.com/cakephp/cakephp/blob/4ac765fdd47d6b185bf9c55dfd5bb068b3daba2c/phpstan.neon#L10) should make phpstan happy.

markstory: Whooops this wasn't supposed to be here.

src/Application.php
markstory: I can split it up if you'd prefer.

othercorey: A simple `foreach (['Cake/Repl', 'Bake'] as $plugin)` should work.

ADmad:
>I don't think it gets a ton of usage by users...

A lot of those developing CMSs would need it, probably more than the repl feature :)

markstory:
@ADmad Fair point I've put mobiledetect back.

config/bootstrap.php
```diff
@@ -165,6 +165,11 @@
 
 /*
  * Setup detectors for mobile and tablet.
+ * To use these detectors run:
+ *
+ * `composer require "mobiledetect/mobiledetectlib:^2.8"`
+ *
+ * Before uncommenting this code.
```
garas, line 172: It is no more commented out.

src/Application.php
```diff
@@ -127,10 +127,12 @@ public function services(ContainerInterface $container): void
...
-        try {
-            $this->addPlugin('Bake');
-        } catch (MissingPluginException $e) {
-            // Do not halt if the plugin is missing
+        foreach (['Cake/Repl', 'Bake'] as $plugin) {
+            try {
+                $this->addPlugin($plugin);
+            } catch (MissingPluginException $e) {
+                // Do not halt if the plugin is missing
+            }
```
raul338, line 135:
```suggestion
        $this->addOptionalPlugin('Cake/Repl');
        $this->addOptionalPlugin('Bake');
```
IMHO Is cleaner

src/Application.php
```diff
@@ -127,10 +127,8 @@ public function services(ContainerInterface $container): void
      */
     protected function bootstrapCli(): void
     {
-        try {
-            $this->addPlugin('Bake');
-        } catch (MissingPluginException $e) {
-            // Do not halt if the plugin is missing
+        $this->addOptionalPlugin('Cake/Repl');
+        $this->addOptionalPlugin('Bake');
         }
```
othercorey, line 132:
```suggestion
        $this->addOptionalPlugin('Cake/Repl');
        $this->addOptionalPlugin('Bake');
```

PR #831, opened 2021-03-18
Enable phpstan analyses on templates files

CauanCabral:
Fix all detected issues

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -2,7 +2,10 @@ parameters:
     level: 7
     checkMissingIterableValueType: false
     treatPhpDocTypesAsCertain: false
+    bootstrapFiles:
+        - config/paths.php
     paths:
         - src
+        - templates
     excludes_analyse:
         - src/Console/Installer.php
--- a/templates/Error/error400.php
+++ b/templates/Error/error400.php
@@ -8,10 +8,10 @@
 use Cake\Core\Configure;
 use Cake\Error\Debugger;
 
-$this->layout = 'error';
+$this->setLayout('error');
 
 if (Configure::read('debug')) :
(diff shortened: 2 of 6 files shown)


ADmad:
Doing static analysis on template files is an overkill.

There are no actual issues in the template code and most of the changes are quite unnecessary.

CauanCabral:
I almost agree, but I think at least [this one](https://github.com/cakephp/app/pull/831/files#diff-1f1107533aee24daa1e811240d4a9f9a702040350a602340ec320ed3eede9046L151) is a bug.

`Exception::getMessage()` should return string, so I think `$errorMsg` is a typo for `$attributes`.

phpstan.neon
```diff
@@ -2,7 +2,10 @@ parameters:
     level: 7
     checkMissingIterableValueType: false
     treatPhpDocTypesAsCertain: false
+    bootstrapFiles:
+        - config/paths.php
     paths:
         - src
+        - templates
```
markstory, line 9: Needing to set `@var` in all the templates is going to be tedious, can we skip this change?

ADmad:
> Exception::getMessage() should return string, so I think $errorMsg is a typo for $attributes.

Then we can fix just that particular line.

In a full app running static analysis on templates will most likely return a lot of false positives, due to lack of context. Having to add `@var` tags for all view variables used in a template would be highly annoying.

dereuromark:
> In a full app running static analysis on templates will most likely return a lot of false positives, due to lack of context. Having to add @var tags for all view variables used in a template would be highly annoying.

IdeHelper helps doing that, but I agree that this should rather stay opt-in for now.

CauanCabral:
Fixed with #832

PR #834, opened 2021-04-09
Fix TypeError when action is called without any arguments.

### LINKED ISSUE #833: TypeError controller
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Application Skeleton Version: latest master

### What you did
Someone accesses Pages as `/pages/`

### What happened

> [TypeError] Argument 1 passed to App\Controller\PagesController::display() must be of the type string, null given, called in /var/www/.../site/vendor/cakephp/cakephp/src/Controller/Controller.php on line 531 in /var/www/.../site/src/Controller/PagesController.php on line 73

No internal server error but normal 404

Looking at the stacktrace we either need to make `public function display(string ...$path): ?Response` nullable
or we need to make sure a (casted) empty string is used instead of null?

ADmad:
Fixes #833.

### DIFF
--- a/src/Controller/PagesController.php
+++ b/src/Controller/PagesController.php
@@ -34,7 +34,7 @@ class PagesController extends AppController
     /**
      * Displays a view
      *
-     * @param string ...$path Path segments.
+     * @param string|null ...$path Path segments.
      * @return \Cake\Http\Response|null
      * @throws \Cake\Http\Exception\ForbiddenException When a directory traversal attempt.
      * @throws \Cake\View\Exception\MissingTemplateException When the view file could not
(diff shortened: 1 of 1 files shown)


othercorey:
Looks ok.

garas:
It is regression in `Cake\Controller\ControllerFactory::invoke()`.

In 4.1, action is invoked with no arguments, in 4.2 with one argument `null`.

othercorey:
> It is regression in `Cake\Controller\ControllerFactory::invoke()`.
> 
> In 4.1, action is invoked with no arguments, in 4.2 with one argument `null`.
> 
> It needs to handle variadic parameters.

Can you open an issue for this?

garas:
I opened new issue at https://github.com/cakephp/cakephp/issues/15441.

markstory:
Seems like we should fix the ControllerFactorynissue instead.

garas:
Fixed in https://github.com/cakephp/cakephp/pull/15445

PR #839, opened 2021-05-08
Add debugkit connection status to welcome page

markstory:
Separate the 'installed' from 'works' states to make setup issues easier to understand.

Screenshot from 2021-05-07 23-46-03
Screenshot from 2021-05-07 23-48-40

### DIFF
--- a/templates/Pages/home.php
+++ b/templates/Pages/home.php
@@ -22,6 +22,25 @@ use Cake\Http\Exception\NotFoundException;
 
 $this->disableAutoLayout();
 
+$checkConnection = function (string $name) {
+    $error = null;
+    $connected = false;
+    try {
+        $connection = ConnectionManager::get($name);
+        $connected = $connection->connect();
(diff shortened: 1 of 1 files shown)


templates/Pages/home.php
```diff
@@ -168,7 +175,15 @@
                         <?php if (Plugin::isLoaded('DebugKit')) : ?>
                             <li class="bullet success">DebugKit is loaded.</li>
                         <?php else : ?>
-                            <li class="bullet problem">DebugKit is NOT loaded. You need to either install pdo_sqlite, or define the "debug_kit" connection name.</li>
+                            <li class="bullet problem">DebugKit is <strong>not</strong> loaded.</li>
+                        <?php endif; ?>
+                        <?php
+                        $result = $checkConnection('debug_kit');
```
garas, line 181: Connection check should be only in `if (Plugin::isLoaded('DebugKit'))` block.

PR #841, opened 2021-06-06
Restore rewriting warning

cleptric:
I could not find any code that would ever display this message


ndm2:
Please [don't kill it off completely](https://github.com/cakephp/app/pull/683/files#diff-06a90010cc659f0efa07d587f7a61a43920979b64f0c285ec535d00684ee32dfL147-L150) 😉 If you could instead add the rules back, that would be nice.

cleptric:
@ndm2 Can you point me to the place in the code that would actually display the hidden div then?

ndm2:
There is no such place, it works the other way around, the linked rule would hide it, and when URL rewriting is misconfigured so that it would cause the CSS file to not be loaded, then that section would be displayed (if it weren't commented out) as there would be no CSS that would hide it.

ADmad:
It's seems the related CSS class in home.css was mistakenly removed during the redesign. So it should be restored and this message uncommented.

cleptric:
@ndm2 @ADmad Got it, thanks. I restored it and added some inline styles to make it pop up better without any css loaded.

PR #843, opened 2021-06-12
Update application skeleton for new fixtures.

markstory:
Using the schema dump file is not ideal. Once cakephp/migrations#488 is merged I will update this to use migrations instead.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -17,7 +17,6 @@ declare(strict_types=1);
 
 use Cake\Core\Configure;
 use Cake\Datasource\ConnectionManager;
-use Cake\TestSuite\Schema\SchemaManager;
 
 /**
  * Test runner bootstrap.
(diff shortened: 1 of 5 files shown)


othercorey:
Looks ok.

tests/bootstrap.php
```diff
@@ -50,3 +50,16 @@
...
 session_id('cli');
+
+// Load schema from a SQL dump file.
+//
+// If your plugin does not use database fixtures you can
+// safely delete this.
+//
+// If you want to support multiple databases, consider
+// using migrations to provide schema for your plugin,
+// and using \Migrations\TestSuite\Migrator to load schema.
```
ADmad, line 61: Would be nice to have a code e.g. for this too.

markstory: I'm going to add it later this week. That will be the 'default' mode for applications as I think re-using migrations is a better dev experience than having to generate a dump file. I was blocked on the migrations changes not being merged.

PR #858, opened 2021-10-02
Update README.md

Varun-Strangelet:
Hello I'm Varun Verma And I'm New in Github Now I'm learning Github I just do Practice of PR request Send Not I do big changes in Readme Doc just add ':' at the end of headings to look good.


othercorey:
We don't put : at the end of headers.

PR #859, opened 2021-10-04
Suggest using the fixture factories in composer.json

pabloelcolombiano:
This is a proposal to suggest using the fixture factories in the app. I am also open to put it in the dev requirements.


saeideng:
you can add the plugin into https://github.com/FriendsOfCake/awesome-cakephp

pabloelcolombiano:
@saeideng this is already done.

dereuromark:
Usually, we just keep and promite the addons/plugins in awesome list until they reach a certain level of "usage".
E.g. >> 100 stars or provide a general functionality that the majority of people will end directly end up using.
I am not sure the factories are there just yet.

saeideng:
there are variety of plugins that users want to use them in special usage
and most plugins do not use by all users

so I think we just could keep them in awesome list and closing this Issue

pabloelcolombiano:
Got your point @dereuromark and @saeideng. I'll get back to it when the package has reached the required amount of audience.

PR #862, opened 2021-10-13
Import Cake\Codeception\Console\Installer

szepeviktor:
Minus one variable 🚀

### DIFF
--- a/src/Console/Installer.php
+++ b/src/Console/Installer.php
@@ -20,6 +20,7 @@ if (!defined('STDIN')) {
     define('STDIN', fopen('php://stdin', 'r'));
 }
 
+use Cake\Codeception\Console\Installer;
 use Cake\Utility\Security;
 use Composer\Script\Event;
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you.

othercorey:
Should we even suggest this anymore?

ADmad:
Probably not, the codeception plugin isn't maintained.

LordSimal:
The class `Cake\Codeception\Console\Installer` doesn't seem to exist for me. I am on CakePHP 4.3.
Or am I missing something?

ndm2:
It's in the [**codeception plugin**](https://github.com/cakephp/codeception/blob/master/src/Console/Installer.php).

LordSimal:
I was just confused to see code/use statements which are not resolvable in the "base" app template because the plugin is not in the composer.json

If this is intended then forget that I asked 😄

PR #863, opened 2021-10-13
Update CI config.

ADmad:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.2', '7.4', '8.0', '8.1']
+        php-version: ['7.2', '7.4', '8.0']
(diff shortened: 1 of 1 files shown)


dereuromark:
composer/packagist is down right now.. we gottoa rerun those.

dereuromark:
OK, now we can see something
```
Deprecated Error: Cake\View\ViewBuilder implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary)
In [/home/runner/work/app/app/vendor/cakephp/cakephp/src/View/ViewBuilder.php, line 37]

PHP Fatal error:  Type of Cake\View\Exception\MissingTemplateException::$file must be string (as in class Exception) in /home/runner/work/app/app/vendor/cakephp/cakephp/src/View/Exception/MissingTemplateException.php on line 23
```

othercorey:
Shouldn't this be for 4.next?

PR #865, opened 2021-10-19
Add bash completion script.

ADmad:

### DIFF
--- a/bin/bash_completion.sh
+++ b/bin/bash_completion.sh
@@ -1,0 +1,47 @@
+#
+# Bash completion file for CakePHP console.
+# Copy this file to a file named `cake` under `/etc/bash_completion.d/`.
+# For more info check https://book.cakephp.org/4/en/console-commands/completion.html#how-to-enable-bash-autocompletion-for-the-cakephp-console
+#
+
+_cake()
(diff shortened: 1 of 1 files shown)


dereuromark:
Does that mean it doesnt need any of the extraction tools/ideas like
https://github.com/dereuromark/cakephp-ide-helper/compare/command-line
to get it working for bash?

Is that true for both commands and shell?

ADmad:
It just works:tm: once you follow https://book.cakephp.org/4/en/console-commands/completion.html#how-to-enable-bash-autocompletion-for-the-cakephp-console

dereuromark:
Yeah, it "kind of" works.
It seems to have some troubling output once you reach the end, instead of just continuing.
But looks like an improvement for sure.
I wonder if we should add a composer script to quickly generate those 3 types: `composer completion` or `composer complete`

ADmad:
> It seems to have some troubling output once you reach the end, instead of just continuing.

Yes, there are some issues with the `CompletionCommand` (or maybe this script) which need to be addressed. 

> I wonder if we should add a composer script to quickly generate those 3 types: composer completion

That would be nice.

PR #873, opened 2021-11-23
Fix .gitignore so potential icon assets are tracked

rmard:
This fixes issue cakephp/cakephp/issues/16138 by using GitHub's .gitignore [template suggestion](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore#L7)


markstory:
Thank you 👍

PR #880, opened 2022-01-13
[4.x] add allow-plugins

saeideng:
also composer will prompt for `"dealerdirect/phpcodesniffer-composer-installer": true`


dereuromark:
> also composer will prompt for "dealerdirect/phpcodesniffer-composer-installer": true

yeah, that should also be added

PR #883, opened 2022-01-25
copy Debugger/editor to app_local.php

saeideng:
this is more a local config
so could be in `app_local.php`

### DIFF
--- a/config/app_local.example.php
+++ b/config/app_local.example.php
@@ -75,6 +75,15 @@ return [
     ],
 
     /*
+     * Debugger configuration
+     *
+     * See app.php for more info.
+     */
+    'Debugger' => [
+        'editor' => 'phpstorm',
+    ],
+
+    /*
      * Email configuration.
      *
      * Host and credential configuration in case you are using SmtpTransport


ndm2:
I'm looking forward to 2030, one fine day this is actually going to work platform wide 😃

https://youtrack.jetbrains.com/issue/IDEA-65879

config/app_local.example.php
```diff
@@ -74,6 +74,15 @@
         ],
     ],
 
+    /*
+     * Debugger configuration
+     *
+     * See app.php for more info.
```
ADmad, line 80: It should be removed from app.php. Having a config at 2 places will just confuses new users.

saeideng: but I prefer to kept it in app.php , because it saves in VCS , and it avoids to forgetting config name

also some of other configs are in 2 places , and no problem
anyway I can comment original code , but removing is not good idea IMO

ADmad: If you are going to keep it in app.php then what's the point in having it in the local config file?

Anyway you are free to do do whatever you want in your app but having the duplication is not a good default for the app skeleton.

dereuromark: @saeideng I think it is pretty clear that every value in app.php can be overwritten in local file
In fact, this will be the case for also many other options as per local dev.
So I guess keeping it in the main file suffices for most - as they know how to copy paste when it becomes necessary.

With this we keep the local file concise and minimal, makes it also easier to review.

saeideng: > what's the point in having it in the local config file?


2 team members could have 2 different editor

saeideng: closing the PR, we want to keep it in VCS

PR #884, opened 2022-02-05
Update application skeleton to use new traps

markstory:
Use the new error and exception traps in new applications.

Refs cakephp/cakephp#16228


othercorey:
CakePHP 4.4 requires PHP 7.4+ now. Fixed.

PR #891, opened 2022-03-24
remove extra `Engine` after class name

### LINKED ISSUE #890: Incorrect cache engine name in home page
This is a (multiple allowed):

* [x] bug

* CakePHP Application Skeleton Version: 4.3.1

### What you did

Open the home page.

### What happened

In Filesystem section:

> The Cake\Cache\Engine\FileEngineEngine is being used for core caching. To change the config edit config/app.php

> The Cake\Cache\Engine\FileEngine is being used for core caching. To change the config edit config/app.php

### Information

This displays `className` of `_cake_core_` cache config, followed by "Engine".
Previously `className` was short name like `File`, but changed to FQCN at 4f7588486e34df7da712449f137fb19dd4000622.
I think it's good to display FQCN even when configured with short name, though it requires instantiating the engine.

saeideng:
Close #890

### DIFF
--- a/templates/Pages/home.php
+++ b/templates/Pages/home.php
@@ -147,7 +147,7 @@ $cakeDescription = 'CakePHP: the rapid development PHP framework';
 
                         <?php $settings = Cache::getConfig('_cake_core_'); ?>
                         <?php if (!empty($settings)) : ?>
-                            <li class="bullet success">The <em><?= $settings['className'] ?>Engine</em> is being used for core caching. To change the config edit config/app.php</li>
+                            <li class="bullet success">The <em><?= $settings['className'] ?></em> is being used for core caching. To change the config edit config/app.php</li>
                         <?php else : ?>
                             <li class="bullet problem">Your cache is NOT working. Please check the settings in config/app.php</li>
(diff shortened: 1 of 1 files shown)


templates/Pages/home.php
```diff
@@ -147,7 +147,7 @@
 
                         <?php $settings = Cache::getConfig('_cake_core_'); ?>
                         <?php if (!empty($settings)) : ?>
-                            <li class="bullet success">The <em><?= $settings['className'] ?>Engine</em> is being used for core caching. To change the config edit config/app.php</li>
+                            <li class="bullet success">The <em><?= $settings['className'] ?></em> is being used for core caching. To change the config edit config/app.php</li>
```
dereuromark, line 150: h() would also be good to add

saeideng: `$settings['className']` comes from config file , not user input

dereuromark: Still good practice imo

markstory: I can add it separately. We should have it if only to model safe template authoring.

PR #895, opened 2022-04-10
4.4 - Allow exception rendering to auto-detect based on environment.

markstory:
Add additional documentation for new options in 4.4

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -166,8 +166,9 @@ return [
      * - `log` - boolean - Whether or not you want exceptions logged.
      * - `exceptionRenderer` - string - The class responsible for rendering
      *   uncaught exceptions. If you choose a custom class you should place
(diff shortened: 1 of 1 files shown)


config/app.php
```diff
@@ -166,8 +166,9 @@
      * - `log` - boolean - Whether or not you want exceptions logged.
      * - `exceptionRenderer` - string - The class responsible for rendering
      *   uncaught exceptions. If you choose a custom class you should place
-     *   the file for that class in src/Error. This class needs to implement a
-     *   render method.
+     *   the file for that class in src/Error. This class needs to implement a `render()` method.
+     *   `errorRenderer` - string - The class responsible for rendering PHP errors. This class needs
```
othercorey: We should mention the custom renderer is used for both cli and non-cli environments.

PR #909, opened 2022-09-12
Update Application.php to remove route cache code comment

umer936:
Also adds note on plugin to add cached routing

Unsure if this PR should be against `4.x` branch or `4.next`. I can resubmit if it's the latter.


othercorey:
The target branch should match the cakephp version it targets. If the target was already released then it should go into 4.x.

umer936:
Then I believe it is correct to go into 4.x as the code change removing the 2nd parameter was previously committed.

PR #916, opened 2022-10-02
4.x - use local google font files instead of external URL

LordSimal:
Refs.: #915

This loads the `Raleway` Google Font locally instead of externally.


dereuromark (approved):
Big yes
Also legal wise

PR #925, opened 2022-12-10
Re-enable migrations plugin

ADmad:


ADmad:
`cs-stan` job is failing because phpstan couldn't be installed `403: rate limit exceeded.

PR #927, opened 2022-12-13
Update minimum PHP version for Cake 5

othercorey:


config/requirements.php
```diff
@@ -20,8 +20,8 @@
 /*
  * You can remove this if you are confident that your PHP version is sufficient.
  */
-if (version_compare(PHP_VERSION, '7.2.0') < 0) {
-    trigger_error('Your PHP version must be equal or higher than 7.2.0 to use CakePHP.', E_USER_ERROR);
+if (version_compare(PHP_VERSION, '8.1.0', '<')) {
```
ADmad, line 23: I don't remember why we needed this, that's composer's job.

dereuromark: Wasnt there still a way to install without composer or alike?

garas: Composer added this check in 2.0.0.

ADmad: > Wasnt there still a way to install without composer or alike?

We do or did full zipballs. I personally have no interest in catering for that tiny fraction of users.

garas: Zipballs already include vendor/ with Composer PHP version check.

PR #928, opened 2022-12-13
Remove runtime check of the PHP version

garas:
PHP version check is now part of Composer


dereuromark:
What if people use composer v1?
Doesnt seem to hurt too much to keep it?

garas:
Composer v1 is EOL few year ago, so we could assume people have already upgraded Composer when trying new CakePHP version, `composer install` would still prevent from installing.

ADmad:
@garas Composer doesn't do the extension checks?

garas:
> @garas Composer doesn't do the extension checks?

Extension checks are added only if `composer.json` has `"config": { "platform-check": true }` (default value `"php-only"`).

So extension checks could be also removed if we add this key to  `composer.json`.

However `lib-*` check is not generated by Composer, so ICU version would be the only checked in this file.

https://getcomposer.org/doc/07-runtime.md#platform-check

ADmad:
> So extension checks could be also removed if we add this key to composer.json.

Then I would prefer we use the feature already provided by composer instead of adding our own checks.

> However lib-* check is not generated by Composer, so ICU version would be the only checked in this file.

The ICU version check could probably be removed. Current ICU version is 71.1 and ICU 50.2 (we check for >50.1) was released on 2019-04-11. It's highly unlikely that an environment with PHP 8.1+ has ICU version older than 50.2.

garas:
I've completely removed `config/requirements.php`.

ICU 50.1 or newer is required by PHP since 7.4.0, so this check is also unneeded.

PR #932, opened 2023-01-05
5.x: fix debug_kit version string

LordSimal:
The cake5 branch name for the debug_kit plugin seems to have changed.
See https://github.com/cakephp/debug_kit/tree/5.x

PR #933, opened 2023-01-05
5.x fix stan

LordSimal:
This PR contains the commit from https://github.com/cakephp/app/pull/932 so CI can actually run.

Due to the recent merge of generics annotations in https://github.com/cakephp/cakephp/pull/16891 this of course introduced some psalm problems here, but nothing too major.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "mobiledetect/mobiledetectlib": "^3.74"
     },
     "require-dev": {
-        "cakephp/bake": "dev-cake5",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -14,7 +14,7 @@
     "require-dev": {
         "cakephp/bake": "dev-cake5",
```
ADmad: Shouldn't this be updated too?

LordSimal: yup, I missed that one 😄 
seems that packagist cached that version string

PR #934, opened 2023-01-06
Move tests.sqlite to tmp

cleptric:
~~The test connection url defaults to `env('DATABASE_TEST_URL', 'sqlite://127.0.0.1/tests.sqlite')`, hence ignoring the generated SQLite file might be useful.~~

Moving the `tests.sqlite` into `tmp/` by default, so it lives next to `debug_kit.sqlite` and is ignored by default.

### DIFF
--- a/config/app_local.example.php
+++ b/config/app_local.example.php
@@ -70,7 +70,7 @@ return [
             'password' => 'secret',
             'database' => 'test_myapp',
             //'schema' => 'myapp',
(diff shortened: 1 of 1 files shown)


dereuromark:
Can't we just put it inside tmp/?
Then we wouldnt need to clutter the root more.

cleptric:
Yes, makes more sense.

PR #937, opened 2023-02-09
update links

Arhell:


othercorey:
We shouldn't modify the external license file.

Arhell (commented):
Done

PR #941, opened 2023-03-01
4.x: remove loading repl plugin in bootstrap cli

LordSimal:
cakephp/repl isn't being installed by default anymore, therefore we shouldn't load it by default as well.
Refs: https://github.com/cakephp/app/pull/878

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -124,7 +124,6 @@ class Application extends BaseApplication
      */
     protected function bootstrapCli(): void
     {
-        $this->addOptionalPlugin('Cake/Repl');
         $this->addOptionalPlugin('Bake');
 
         $this->addPlugin('Migrations');


othercorey:
Isn't that what addOptionalPlugin is for?

garas:
> Isn't that what addOptionalPlugin is for?

It is for dev dependencies (`require-dev` in `composer.json`).

Repl in not included neither in `require` or `require-dev`, only in `suggest`.

othercorey:
Sure, but it's the same concept. We could just comment it out I guess.

LordSimal:
> Sure, but it's the same concept. We could just comment it out I guess.

By that logic we should then optionally load 
- `markstory/asset_compress` and
- `dereuromark/cakephp-ide-helper`

as well

othercorey:
Sure. If we're suggesting it. This is just a helpful template to show where things go, right.

LordSimal:
I am fine with either one of the solutions - either remove all or add all suggestions which are actual CakePHP plugin suggestions

othercorey:
@ADmad what do you think? I don't have a strong opinion, just thought it could be used for this.

ADmad:
I don't have a strong opinion on this either but would lean towards removing it. I don't remember anyone mentioning using the repl plugin in any of our help channels in a long time.

ADmad:
Also one major gripe I have with the repl plugin / psysh is that it pulls in a symfony package which override our `dd()` function.

othercorey:
Sounds good, merging since this isn't a major issue.

markstory:
> either remove all or add all suggestions which are actual CakePHP plugin suggestions

Removing makes sense to me. We don't need to clutter people's applications with a bunch of commented out code when we also give them CLI tools to add the lines back in if required.

PR #943, opened 2023-03-11
Configuring Docker

JulyanoF:
Configuring minumum requirements to run a CakePHP application with Docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM php:7.4-apache-buster
+
+# Installing minimum system packages
+RUN apt-get update && apt-get install -y \
+    libicu-dev \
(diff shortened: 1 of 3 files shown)


ADmad:
The environment needs can vary a lot for each app/individual. For e.g. mysql vs postgres, nginx vs apache etc. 

There are already other generic LAMP VMs available which can be used for Cake apps, so these files dont need to be part of this repo.

markstory:
> There are already other generic LAMP VMs available which can be used for Cake apps, so these files dont need to be part of this repo.

This is my concern as well. Should we link to a recommended image instead? What are other frameworks like symfony, laravel and laminas doing for this?

PR #948, opened 2023-04-13
4.next: add example on how to set custom date/time format

LordSimal:
Users often don't know where/how to set a custom date/time format - as well as to use the ICU format instead of the PHP format.

Should we also add the mutable classes as well? I personall don't think so.

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -221,3 +221,9 @@ TypeFactory::map('time', StringType::class);
 //Inflector::rules('plural', ['/^(inflect)or$/i' => '\1ables']);
 //Inflector::rules('irregular', ['red' => 'redlings']);
 //Inflector::rules('uninflected', ['dontinflectme']);
+
+// set a custom date and time format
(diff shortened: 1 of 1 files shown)


markstory:
We're phasing mutable classes out so they don't need to be in the skeleton.

config/bootstrap.php
```diff
@@ -221,3 +221,9 @@
 //Inflector::rules('plural', ['/^(inflect)or$/i' => '\1ables']);
 //Inflector::rules('irregular', ['red' => 'redlings']);
 //Inflector::rules('uninflected', ['dontinflectme']);
+
+// set a custom date and time format
+// see https://book.cakephp.org/4/en/core-libraries/time.html#setting-the-default-locale-and-format-string
+// and https://unicode-org.github.io/icu/userguide/format_parse/datetime/#datetime-format-syntax
```
othercorey, line 227: I'd prefer a block comment here so it is separate from the code example

PR #951, opened 2023-05-31
5.x: upgrade to PHPUnit 10

LordSimal:
Refs: https://github.com/cakephp/cakephp/pull/16992

Just noticed there is a little bug in the debug_kit which was introduced when upgrading to PHPUnit 10

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,7 @@
 .Spotlight-V100
 .Trashes
 # Icon must end with two \r
-Icon
-
+Icon
 ehthumbs.db
 Thumbs.db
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -14,15 +14,16 @@
 .Spotlight-V100
 .Trashes
 # Icon must end with two \r
-Icon
+Icon
+
```
garas: Revert this line, it must end with two \r

LordSimal: Lovely that my editor seems to do that automatically...

PR #953, opened 2023-06-25
Use Debugger::trimPath() and Debugger::editorUrl() for error templates

andrii-pukhalevych:
Use Debugger for error templates + align error400 with error500


markstory:
Thank you 🎉

PR #955, opened 2023-07-16
5.x: remove static method in routes.php

LordSimal:
This is my proposed version of how the routes.php should look like.

Why?

Depending on the application one has to build middlewares inside the routes.php
Some middlewares need the application instance like our own [AuthenticationMiddleware](https://github.com/cakephp/authentication/blob/3.x/src/Middleware/AuthenticationMiddleware.php#L50)

E.g. having only some routes which need authentication and not all routes.

But accessing `$this` inside a static function is not allowed. Passing on `$this` via 
```
return static function (RouteBuilder $routes) use ($this)
``` 
is also not possible.

So one would have to build the objects outside of the static method and pass them through, which is kind of bad imho.

If we only added that static method so IDE's can have auto-completetion for the available RouteBuilder methods this is in my opinion the better solution.

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -21,71 +21,73 @@
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
 
+/**
+ * @var \Cake\Routing\RouteBuilder $routes
+ */
+
 use Cake\Routing\Route\DashedRoute;
 use Cake\Routing\RouteBuilder;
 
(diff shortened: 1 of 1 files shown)


ADmad:
The problem you described can be avoided by just removing the `static` keyword for the function, there's no need to remove the anonymous function all together.

PR #956, opened 2023-07-16
Remove `static` keyword from the function in routes.php.

ADmad:
This allows using `$this` in the function to reference the application class instance. This is required for cases like building an AuthencationMiddleware instance.

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -24,7 +24,11 @@
 use Cake\Routing\Route\DashedRoute;
 use Cake\Routing\RouteBuilder;
 
-return static function (RouteBuilder $routes) {
+/**
+ * This file is loaded in the context of the `Application` class. So you can
+ * `$this` to reference the application class instance if required.
+ */
+return function (RouteBuilder $routes) {
     /*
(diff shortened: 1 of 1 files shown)


config/routes.php
```diff
@@ -24,7 +24,11 @@
 use Cake\Routing\Route\DashedRoute;
 use Cake\Routing\RouteBuilder;
 
-return static function (RouteBuilder $routes) {
+/**
```
ravage84: I think this should only be a simple multi-line comment:

```suggestion
/*
```

config/routes.php
```diff
@@ -24,7 +24,11 @@
...
+ * This file is loaded in the context of the `Application` class. So you can
+ * `$this` to reference the application class instance if required.
```
ravage84: Missing "use"?
```suggestion
 * This file is loaded in the context of the `Application` class.
  * So you can use  `$this` to reference the application class instance
  * if required.
```

ravage84:
When testing two of our CakePHP 4.4 applications, they run through the test suite with or without. 👍

PR #969, opened 2023-09-13
Update AjaxView.php use setResponse

markusramsak:


ADmad:
This is an unnecessary change (not to mention fails even the linter). There's no problem assigning the protected property directly.

markusramsak:
> This is an unnecessary change (not to mention fails even the linter). There's no problem assigning the protected property directly.

I made a typo "=" should of course be ";"
I know that assigning a protected property is not an error but I thought that we could use setResponse and setRequest everywhere without exceptions to be consistent ...

PR #970, opened 2023-09-14
Update error500.php cannot use protected property

markusramsak:
cannot use protected property `params` -> use `->getBoundParams()`

### DIFF
--- a/templates/Error/error500.php
+++ b/templates/Error/error500.php
@@ -24,9 +24,9 @@ if (Configure::read('debug')) :
         <?= h($error->queryString) ?>
     </p>
 <?php endif; ?>
-<?php if (!empty($error->params)) : ?>
+<?php if (!empty($error->getBoundParams())) : ?>
     <strong>SQL Query Params: </strong>
-    <?php Debugger::dump($error->params) ?>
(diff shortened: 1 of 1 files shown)


othercorey:
The annotation seems incorrect here if $error can be an instance of `Error`.

templates/Error/error500.php
```diff
@@ -24,9 +24,9 @@
         <?= h($error->queryString) ?>
     </p>
 <?php endif; ?>
-<?php if (!empty($error->params)) : ?>
+<?php if (!empty($error->getBoundParams())) : ?>
```
othercorey, line 27: If $error can be multiple classes, should probably wrap all of this in a instanceof check.

markusramsak: but the Error class also only has getBoundParams() and protected params ...

markstory: Which exception class has `getBoundParams()`? I only see it on `Database\Statement\Statement`.

ADmad: This now needs to be a `method_exists()` check now or better yet an `instanceof` check for specific exception class. Suggested change will generate errors for other error classes.

PR #971, opened 2023-09-15
5.x: use plugin loading via configuration array

LordSimal:
I really like the new plugin loading functionality via `config/plugins.php`. 

This allows the Application class to be pretty clean.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "license": "MIT",
     "require": {
         "php": ">=8.1",
-        "cakephp/cakephp": "^5.0.0",
+        "cakephp/cakephp": "^5.0.1",
         "cakephp/migrations": "^4.0.0",
         "cakephp/plugin-installer": "^2.0",
         "mobiledetect/mobiledetectlib": "^3.74"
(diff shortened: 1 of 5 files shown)


LordSimal:
Seems weird that the DebugKit is not being loaded in the tests...

LordSimal:
Ah, we do a `include_once` [here](https://github.com/cakephp/cakephp/blob/5.x/src/Http/BaseApplication.php#L186) but these tests require that this file can be loaded multiple times in the same PHP process.

ADmad:
Thanks, this was my final goal, just never got around to finishing it :)

I eventually want plugin installing and enabling to be a single command process, not the 2 steps of `composer install` and `bin/cake plugin load`.

LordSimal:
> I eventually want plugin installing and enabling to be a single command process

yes, that would be the best scenario

markstory:
I like the approach of using `config/plugins.php` more consistently :+1:

LordSimal:
To get 8.1 lowest working here we would have to adjust the composer.json to `"cakephp/cakephp": "^5.0.1",`

PR #975, opened 2023-10-04
Ignore return type hints in `src/Command`

cleptric:
PHPCS will complain about missing return-type hints when you bake your first command. 

```php
Method \App\Command\AppCommand::execute() does not have native return type hint for its return value but it should be possible to add it based on @return
    |       |     annotation "int|null|void".
```

PHPCBF then adds `: int|null|null` 🤔

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -4,6 +4,7 @@
 
     <rule ref="CakePHP"/>
     <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint">
+        <exclude-pattern>*/src/Command/*</exclude-pattern>
         <exclude-pattern>*/src/Controller/*</exclude-pattern>
     </rule>
 


phpcs.xml
```diff
@@ -4,6 +4,7 @@
 
     <rule ref="CakePHP"/>
     <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint">
+        <exclude-pattern>*/src/Command/*</exclude-pattern>
```
dereuromark, line 7: You cannot do this

cleptric: Keen to explain why? It seems to work fine in my apps.

cleptric: I guess as all methods would now be ignored. So maybe we just target `execute()`.

cleptric:
Moved to https://github.com/cakephp/bake/pull/950

Issue #979, opened 2023-11-10
Global function usage only in error with phpstan but not in production/unit test despite migration doc


liqueurdetoile:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Application Skeleton Version: 5.0.1
* Platform and Target: PHP 8.2.11

Hey there,

Just starting with a fresh 5x app install, I was facing a phpstan error with the global translation function `__` : 

```
--------------------------------------------------------------------- 
  Line   Model\Table\WebsitesTable.php
 ------ ---------------------------------------------------------------------
  :96    Function __ not found.
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols
 ------ ---------------------------------------------------------------------
```
Curiously my unit tests were working fine and not complaining about global `__` use. Wandering why, I've found out that my default app's `config/bootstrap.php` still has `require CAKE . 'functions.php';`.

I've read once (and forgot till now) in the migration guide that 

> Global functions are now opt-in. If your application uses global function aliases be sure to add require CAKE . 'functions.php' to you application’s config/bootstrap.php.

We're in the middle of the river with current setup 😬 Either, the default bootstrap should be updated (and my unit tests would have quick failed), either, default phpstan config should be updated :

```yaml
parameters:
    level: 8
    treatPhpDocTypesAsCertain: false
    checkGenericClassInNonGenericObjectType: false
    # add this one
    bootstrapFiles:
        - vendor/cakephp/cakephp/src/functions.php
    paths:
        - src/
```

Finally, it's also possible to get rid of these globals as they are now defined in namespace. For migration ease, they can be split in an optional compatibility plugin so [it won't lure IDE with *unrequired* globals if typing `__(` too  quickly](https://i.postimg.cc/59Rs723v/capture.png) if plugin is not installed.

It's a really quick and naive approach and I'm certainly missing many elements to say what is the best option. If using an IDE, last one seems to be the best though.

By the way, you've made a hard work with CakePHP 5.x. Thanks !

markstory:
> We're in the middle of the river with current setup 😬 Either, the default bootstrap should be updated (and my unit tests would have quick failed), either, default phpstan config should be updated :

Updating the phpstan configuration seems like the lowest friction change here. Removing `functions.php` from application startup is another viable option. I don't prefer that approach because we still have a lot of documentation and examples that are using global functions and I'd prefer to not create confusion for new folks.

[...]

PR #980, opened 2023-11-10
Update phpstan default configuration

liqueurdetoile:
Related to https://github.com/cakephp/app/issues/979

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -2,5 +2,7 @@ parameters:
     level: 8
     treatPhpDocTypesAsCertain: false
     checkGenericClassInNonGenericObjectType: false
+    bootstrapFiles:
+        - vendor/cakephp/cakephp/src/functions.php
     paths:
         - src/
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,6 +3,6 @@ parameters:
     treatPhpDocTypesAsCertain: false
     checkGenericClassInNonGenericObjectType: false
     bootstrapFiles:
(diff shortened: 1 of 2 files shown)


phpstan.neon
```diff
@@ -2,5 +2,7 @@ parameters:
     level: 8
     treatPhpDocTypesAsCertain: false
     checkGenericClassInNonGenericObjectType: false
+    bootstrapFiles:
+        - vendor/cakephp/cakephp/src/functions.php
```
ADmad: I am wondering if it would be better to just include `config/bootstrap.php` instead.

liqueurdetoile: I've dig a bit more. In fact, CakePHP constants are still a problem with my current proposal because :

- Global functions are defined in `vendor/cakephp/cakephp/src/functions.php`, called by `config/bootstrap.php`
- Global constants are defined in `config/paths.php` called by `config/bootstrap.php`

So, I think you're right.

liqueurdetoile: `config/bootstrap.php` will also likely be used to load some dependencies outside composer.

liqueurdetoile (commented):
Damnit. Vscode changed all layout. Is it a problem ?

liqueurdetoile:
I've restored previous layout

LordSimal:
I have the same in my apps, so LGTM 👍🏻

ADmad (approved):

markstory:
Thank you 🎉

PR #984, opened 2023-11-29
Chronos setTestNow for apps test harness default

dereuromark:
Shouldnt we include that as best practice in general?


markstory:
> Shouldnt we include that as best practice in general?

I agree, I've had to do this in a few times.

PR #991, opened 2024-01-03
5.x: add PHP 8.3 to CI

LordSimal:


.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['8.1', '8.2']
+        php-version: ['8.1', '8.2', '8.3']
```
dereuromark: We can replace

LordSimal: ok 🤷🏻

PR #993, opened 2024-02-21
Clean up error templates.

### LINKED ISSUE #992: Error templates
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Application Skeleton Version: 5.0.4.

### What happened
Error templates use `$error->params` which are protected. It would be better to use `$error->errorInfo` maybe...

templates/Error/error400.php line 27, 29
templates/Error/error400.php line 27, 29

ADmad:
With debug enabled the WebExceptionRenderer switches the template to pdo_error.php in case of a PDOException, so having checks for SQL query related stuff in debug mode is redundant.

Refs #970, closes #992.

### DIFF
--- a/templates/Error/error400.php
+++ b/templates/Error/error400.php
@@ -1,12 +1,10 @@
 <?php
 /**
  * @var \App\View\AppView $this
- * @var \Cake\Database\StatementInterface $error
  * @var string $message
  * @var string $url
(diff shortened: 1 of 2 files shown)


othercorey:
Looks ok to me.

PR #998, opened 2024-03-28
Fix missing return type check in ErrorController.php

umer936:
Fixes: 

> ( ! ) Fatal error: Declaration of App\Controller\ErrorController::beforeFilter(Cake\Event\EventInterface $event) must be compatible with App\Controller\AppController::beforeFilter(Cake\Event\EventInterface $event): void in /web/hpca/src/Controller/ErrorController.php on line 44

### DIFF
--- a/src/Controller/ErrorController.php
+++ b/src/Controller/ErrorController.php
@@ -39,7 +39,7 @@ class ErrorController extends AppController
      * beforeFilter callback.
      *
      * @param \Cake\Event\EventInterface<\Cake\Controller\Controller> $event Event.
-     * @return \Cake\Http\Response|null|void
+     * @return void
      */
     public function beforeFilter(EventInterface $event): void
     {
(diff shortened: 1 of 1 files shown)


dereuromark:
Is that fatal saying twice the same exact line?
Weird

But either way, this is actually the correct the way it was so far.

The code does not contain the beforeFilter()
https://github.com/cakephp/app/blob/5.x/src/Controller/AppController.php
And the error controller defines it without actual type as `@return \Cake\Http\Response|null|void`

You added that type yourself, and as per documentation it can return (and in some cases must) Response, e.g. for redirects.
So the typehint `\Cake\Http\Response|null|void` documented is correct.
It would be the actual type `\Cake\Http\Response|null` if you always return something, that's a choice you can make on your project level.

The changes proposed as such can not be adapted/merged here.

ADmad:
Closing for reasons stated above.

PR #999, opened 2024-04-05
5.next: Allow DebugKit config to be set in environment variables

ajibarra:
Please refer to: https://github.com/cakephp/debug_kit/issues/997


dereuromark:
Imo this can directly go into 5.x as those are just skeleton changes for New apps

ajibarra:
Sure I can prepare another PR later today!

PR #1008, opened 2024-07-25
Remove not needed cloaking check

dereuromark:
My CS was noting that


config/bootstrap.php
```diff
@@ -155,7 +155,7 @@
     }
 
     $httpHost = env('HTTP_HOST');
-    if (isset($httpHost)) {
+    if ($httpHost !== null) {
```
dereuromark: I wonder if this should even be 
```suggestion
    if ($httpHost) {
```
As empty string would also be weird here

PR #1011, opened 2024-07-26
Utf8mb4 by default

dereuromark:
Is there any need these days to have to manually toggle this on every time? Or can it be the default now for all dbs?

Refs https://github.com/ddev/ddev/issues/6428#issuecomment-2253037736


ADmad (approved):

othercorey:
Is this still not the default in mysql8?

PR #1019, opened 2024-09-18
Avoid unintentional jump to new core minor.

ADmad:
It often makes apps go kaboom without taking appropriate upgradation steps.


dereuromark:
You can also use ~

PR #1026, opened 2024-11-06
Update home.php

luizcmarin:
mcrypt was deprecated in PHP 7.2.

https://www.php.net/manual/en/intro.mcrypt.php


LordSimal:
thanks 👍🏻

PR #1027, opened 2024-12-13
merge 5.x => 5.next

LordSimal:


composer.json
```diff
@@ -6,7 +6,7 @@
     "homepage": "https://cakephp.org",
     "require": {
         "php": ">=8.1",
-        "cakephp/cakephp": "dev-5.next as 5.1.0",
+        "cakephp/cakephp": "5.1.*",
```
ADmad, line 9: This needs to be set back to 5.next post merge

PR #1045, opened 2025-06-26
Stricter future proof event callback signature.

dereuromark:
Can we do this for the skeleton for new apps?

Refs
https://github.com/cakephp/cakephp/issues/18764

### DIFF
--- a/src/Controller/ErrorController.php
+++ b/src/Controller/ErrorController.php
@@ -41,7 +41,7 @@ class ErrorController extends AppController
      * @param \Cake\Event\EventInterface<\Cake\Controller\Controller> $event Event.
      * @return void
      */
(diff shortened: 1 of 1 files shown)


mirko-pagliai:
But I don't see any problems at the moment.
No method currently explicitly reports `void`.
The user mentions `AppController::beforeFilter()`, but that's not explicit either.

If we want to proceed in this direction, it should be made explicit everywhere (all events of all classes that extend `Controller`) ???

dereuromark:
This will happen, but for BC reasons we cannot change the core (in a patch) release. Generated code and this boilerplate code can be adjusted safely usually.

ADmad (approved):

markstory:
Yes, this is totally fine to do in the skeleton.

PR #1046, opened 2025-06-27
View property "layout" is protected, so use setLayout() instead.

alain-guidon:

### DIFF
--- a/templates/Error/error400.php
+++ b/templates/Error/error400.php
@@ -6,10 +6,10 @@
  */
 use Cake\Core\Configure;
 
-$this->layout = 'error';
+$this->setLayout('error');
 
 if (Configure::read('debug')) :
-    $this->layout = 'dev_error';
+    $this->setLayout('dev_error');
 
     $this->assign('title', $message);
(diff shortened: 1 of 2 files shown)


templates/Error/error400.php
```diff
@@ -6,10 +6,10 @@
  */
 use Cake\Core\Configure;
 
-$this->layout = 'error';
+$this->setLayout('error');
```
dereuromark, line 9: Protected means outside $this context.
But all of the template are running inside, see also the docblock above :)

So none of those changes are necessary.

alain-guidon: Yes... You're right. :-)

I followed the advice given by the code analyzer (VSCode + DEVSENSE PHP) without thinking about the context.
Sorry.

dereuromark: If they report this despite the docblock at the top, maybe you can open a bug ticket with them instead.

alain-guidon: Yeap.
You should close this pull request.
Thanks for your replies.

PR #1047, opened 2025-07-08
clear up table class fallback usage

LordSimal:


dereuromark:
Would be nice if for tests it would behave as it would for non cli.
I wonder if we could set this based on bake running specifically.

PR #1048, opened 2025-07-14
Move table fallback to Bake

dereuromark:
Completes https://github.com/cakephp/bake/pull/1046

As Bake adds this if needed, the behavior should be BC once people use 3.next as minimum for bake.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -48,12 +48,8 @@ class Application extends BaseApplication
         // Call parent to load bootstrap from files.
         parent::bootstrap();
 
-        if (PHP_SAPI !== 'cli') {
-            FactoryLocator::add(
-                'Table',
-                (new TableLocator())->allowFallbackClass(false),
-            );
-        }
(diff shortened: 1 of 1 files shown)


src/Application.php
```diff
@@ -48,12 +46,11 @@ public function bootstrap(): void
...
-            FactoryLocator::add(
-                'Table',
-                (new TableLocator())->allowFallbackClass(false),
-            );
-        }
+        // Use if you need fallback tables
+        //\Cake\Datasource\FactoryLocator::add(
+        //    'Table',
+        //    (new \Cake\ORM\Locator\TableLocator())->allowFallbackClass(false),
+        //);
```
ADmad:
```suggestion
        FactoryLocator::add('Table', (new TableLocator())->allowFallbackClass(false));
```

This call must rename to disable fallbacks, only the `if` block needs to be removed. The default behavior is to auto use `Table` class instance if a class is not found for a model, which we **do not** want.

PR #1050, opened 2025-07-22
Update AppView.php

dereuromark:
`@link` should be used for linking things together. Here, a simple `@see` is correct IMO.

Refs https://github.com/dereuromark/cakephp-ide-helper/blob/b6237cc03a50735bd43a5290e0a9e77972864208/config/app.example.php#L34

### DIFF
--- a/src/View/AppView.php
+++ b/src/View/AppView.php
@@ -22,7 +22,7 @@ use Cake\View\View;
  *
  * Your application's default view class
  *
- * @link https://book.cakephp.org/5/en/views.html#the-app-view
+ * @see https://book.cakephp.org/5/en/views.html#the-app-view
  */
 class AppView extends View
 {


LordSimal:
@see can be a URL or a FQSEN https://docs.phpdoc.org/guide/references/phpdoc/tags/see.html
@link must be a URL https://docs.phpdoc.org/guide/references/phpdoc/tags/link.html

So in this case @link is fine imho since its more specific.

dereuromark:
No, the semantic meaning is different, and also link can and should be used for linking things together IMO.
See referenced PR/changes.

`@link` would mess with the annotator since using link over uses (since uses is used by phpunit for coverage unfortunately).

See should be enough here, the fact that it can be FQCN and referencing internal files and methods does not disqualify its semantic use here.

dereuromark:
Also
https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.see.pkg.html

"Display a link to the documentation for an element"

> The `@see` tag may be used to document any element (global variable, include, page, class, function, define, method, variable)

IMO exactly what we want in these cases

ADmad:
> No, the semantic meaning is different, and also link can and should be used for linking things together IMO.

Which documentation says that?

This is what is says for `@link` from the docs you linked https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.link.pkg.html

> Display a hyperlink to a URL in the documentation
> The guidelines for giving the element name to @link are the same as outlined in the @see manual page.

So I don't see anything wrong with the current `@link` here.

Phpunit may have hijacked this tag for it's own use. Also usage of phpdoc annotations is now deprecated in phpunit and it uses PHP attributes instead.

dereuromark:
Well, the IDEHelper now has to use `@link` either way for linking code and test code - see the linked reasons above.
So I would prefer if we used the `@see` tag for normal doc linking which is reasonable enough IMO.

dereuromark:
I can add https://github.com/dereuromark/cakephp-ide-helper/pull/392 for now, but would prefer if we used a consistent and helpful tag here, so that `@link` in IDE actually is able to also show the linked parts of the system.

dereuromark:
The core seems to also use both randomly.
I would prefer if we consolidated this to prefer link for linking:

    @link \Cake\Core\ObjectRegistry::unload()

and for doc links:

    @see http....
    

also in core code.
I can make a PR here for either direction.

PR #1055, opened 2025-08-23
Use public/ instead of webroot/

### LINKED ISSUE #1054: RFC: 6.x: Use public/ instead of webroot/
I think with v6 we should finally fix this up to be consistent with all other frameworks and the intuitive understanding of
- whats actually "public" from the outside, accessible directly
- and whats "under the hood"

This would make CakePHP compatible with
https://github.com/php-pds/skeleton

Symfony, Laravel and the other major PHP frameworks all use this for years, and it also makes the transition from them to Cake easier, as well as provide less hacks needed for dev tools (e.g. ddev needs to customize this internally to fit Cake).

arshidkv12:
Resolves https://github.com/cakephp/app/issues/1054


dereuromark:
This PR has to wait until we have a 6.x branch.

arshidkv12:
Ok

markstory:
Why are the `webroot` files still included?

> This PR has to wait until we have a 6.x branch.

Why? Application skeletons are generally used for 'new' applications, and we could switch to `public` in a minor release. One potential issue I could see is that plugins will still be using `webroot` and not public.

dereuromark:
I wouldnt want to make such a drastic chance between majors. It only confused everyone and many plugins or alike might expect a certain path.

ADmad:
I wouldn't make this change in a minor either, it would create a lot of confusion and documentation inconsistency.

PR #1069, opened 2025-11-14
Remove extra aliasing

dereuromark:
Do we need those with 

    ConnectionHelper::addTestAliases();

in there?


LordSimal:
Guess this isn't needed anymore

PR #1071, opened 2025-11-24
Security: Prevent Host Header Injection attacks

dereuromark:
## Summary

This PR fixes a **critical security vulnerability** that allows Host Header Injection attacks, which can be exploited to hijack password reset tokens and compromise user accounts.

### The Vulnerability

The current implementation in `config/bootstrap.php` (lines 168-170) dynamically sets `App.fullBaseUrl` from the `HTTP_HOST` header when not configured:

```php
$httpHost = env('HTTP_HOST');
if ($httpHost) {
    $fullBaseUrl = 'http' . $s . '://' . $httpHost;  // ⚠️ Trusts attacker input!
}
```

### Attack Scenario

1. Attacker sends password reset request with malicious Host header: `Host: attacker.com`
2. Application generates reset URL: `http://attacker.com/users/reset_password/valid-token-123`
3. Victim receives email and clicks the malicious link
4. Attacker captures the valid reset token from their server logs
5. Attacker uses stolen token on legitimate domain to reset victim's password → **Account takeover**

### The Fix

**1. Changed `config/app.php`:**
- Set `App.fullBaseUrl` to use `APP_FULL_BASE_URL` environment variable (instead of `false`)
- Enhanced security documentation explaining the risk
- Added clear configuration instructions

**2. Enhanced `config/bootstrap.php`:**
- **Production mode**: Throws exception if `App.fullBaseUrl` is not configured
- **Development mode**: Keeps HTTP_HOST fallback for convenience (with explicit warning)
- Added comprehensive security comments

**3. Updated `config/.env.example`:**
- Added `APP_FULL_BASE_URL` configuration with security documentation
- Provides example value for developers

### Development
✅ **No breaking changes** - HTTP_HOST fallback still works in debug mode

### Production  
⚠️ **Applications MUST configure APP_FULL_BASE_URL** or will fail with clear error:
```
SECURITY: App.fullBaseUrl is not configured. 
This is required in production to prevent Host Header Injection attacks.
Set APP_FULL_BASE_URL environment variable or configure App.fullBaseUrl in config/app.php
```

This is intentional to force proper security configuration in production environments.

As this only applies to new apps, this is BC. current apps are most likely vulnerable in many cases, though.
Maybe debug kit or sth could warn here.

## Configuration Options

Developers can configure this in multiple ways:

**Option 1: Environment Variable (Recommended)**
```bash
APP_FULL_BASE_URL=https://yourdomain.com
```

**Option 2: config/app(_local).php**
```php
'App' => [
    'fullBaseUrl' => 'https://yourdomain.com',
]
```

## Testing

Manual testing performed:
- ✅ Development mode with debug=true: Works as before (uses HTTP_HOST)
- ✅ Production mode without config: Throws clear security exception
- ✅ Production mode with config: Works correctly with configured URL

## References

- [OWASP: Host Header Injection](https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/07-Input_Validation_Testing/17-Testing_for_Host_Header_Injection)
- [PortSwigger: Host Header Attacks](https://portswigger.net/web-security/host-header)
- [CakePHP Documentation: fullBaseUrl](https://book.cakephp.org/5/en/development/configuration.html)

## Related Security Concern

The existing comment in `config/app.php` (line 42) already mentions "if you are concerned about people manipulating the Host header" but didn't enforce it. This PR makes that concern actionable by requiring explicit configuration in production.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -59,7 +59,7 @@ return [
         'webroot' => 'webroot',
         'wwwRoot' => WWW_ROOT,
         //'baseUrl' => env('SCRIPT_NAME'),
-        'fullBaseUrl' => env('APP_FULL_BASE_URL'),
+        'fullBaseUrl' => env('APP_FULL_BASE_URL', false),
         'imageBaseUrl' => 'img/',
         'cssBaseUrl' => 'css/',
         'jsBaseUrl' => 'js/',
--- a/config/app.php
+++ b/config/app.php
@@ -36,10 +36,12 @@ return [
      *      /.htaccess
      *      /webroot/.htaccess
      *   And uncomment the baseUrl key below.
-     * - fullBaseUrl - A base URL to use for absolute links. When set to false (default)
-     *   CakePHP generates required value based on `HTTP_HOST` environment variable.
-     *   However, you can define it manually to optimize performance or if you
-     *   are concerned about people manipulating the `Host` header.
+     * - fullBaseUrl - SECURITY: A base URL to use for absolute links.
+     *   IMPORTANT: This MUST be set in production to prevent Host Header Injection attacks
+     *   that can compromise password reset and other security-critical features.
+     *   Set this via APP_FULL_BASE_URL environment variable or directly in config.
+     *   Example: 'https://yourdomain.com'
+     *   When not set, the application will throw an exception in production mode.
      * - imageBaseUrl - Web path to the public images/ directory under webroot.
      * - cssBaseUrl - Web path to the public css/ directory under webroot.
      * - jsBaseUrl - Web path to the public js/ directory under webroot.
@@ -57,7 +59,7 @@ return [
         'webroot' => 'webroot',
         'wwwRoot' => WWW_ROOT,
         //'baseUrl' => env('SCRIPT_NAME'),
-        'fullBaseUrl' => false,
+        'fullBaseUrl' => env('APP_FULL_BASE_URL'),
         'imageBaseUrl' => 'img/',
         'cssBaseUrl' => 'css/',
         'jsBaseUrl' => 'js/',
--- a/config/.env.example
+++ b/config/.env.example
@@ -18,6 +18,9 @@ export DEBUG="true"
 export APP_ENCODING="UTF-8"
 export APP_DEFAULT_LOCALE="en_US"
 export APP_DEFAULT_TIMEZONE="UTC"
+# SECURITY: Set this to your domain to prevent Host Header Injection attacks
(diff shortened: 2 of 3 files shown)


config/app.php
```diff
@@ -57,7 +59,7 @@
         'webroot' => 'webroot',
         'wwwRoot' => WWW_ROOT,
         //'baseUrl' => env('SCRIPT_NAME'),
-        'fullBaseUrl' => false,
+        'fullBaseUrl' => env('APP_FULL_BASE_URL'),
```
ADmad: Wouldn't removing the `false` break local dev setups?

dereuromark: So we keep false here then as default?

LordSimal:
This is a bad idea as this generates a blank page without any error for `debug => false` in the default setup as its before ErrorHandler middleware has been set up + no log message in logs/error.log as `Log::setConfig()` is called after this exception.

LordSimal:
support chat also asked how this security issue works with a multi domain setup

sure one can dynamically create a variable before the return array in e.g. `config/app_local.php` and use that as `fullBaseUrl` but it would be better if we can come up with a separate array based config key where all allowed HTTP_HOST values are configured and only if the current domain doesn't match that we handle that.

ADmad:
>  ..as its before ErrorHandler middleware has been set up..

Exceptions generated before the middleware queue is executed are expected to be handled by the exception trap registered here
https://github.com/cakephp/app/blob/5.x/config/bootstrap.php#L129

LordSimal:
sure, they are handled there, but the only thing thats being done there is trying to log the exception which it doesn't know where, as the Log config has not yet been set
https://github.com/cakephp/app/blob/5.x/config/bootstrap.php#L203

LordSimal:
The WebException renderer also doesn't work as the cache config required for our I18n methods also doesn't exist yet
https://github.com/cakephp/cakephp/blob/5.x/src/Error/Renderer/WebExceptionRenderer.php#L355
https://github.com/cakephp/app/blob/5.x/config/bootstrap.php#L199

LordSimal:
I'd personally move that check logic into a middleware instead of doing that inside bootstrap.

with that it only applies to HTTP requests and doesn't requirey any custom fiddling around for CLI tools or testing + it happens after everything has been setup and therefore our error handling logic works as well

ADmad:
In production mode could we perhaps check for `env('SSL_TLS_SNI')` if `'App.fullBaseUrl'` is not set?

PR #1076, opened 2026-01-17
Move Host Header Injection check to middleware

dereuromark:
## Summary

This PR addresses the concerns raised in #1071 regarding the Host Header Injection security fix. The original implementation threw an exception in `bootstrap.php` before error handling was initialized, causing:

- Blank pages in production (no error rendering)
- No logging (Log::setConfig() hadn't run yet)
- Issues with cache-based error rendering

This PR moves the security check to a dedicated `HostHeaderMiddleware` that runs after `ErrorHandlerMiddleware`, ensuring proper error handling.

### Changes

- **New `HostHeaderMiddleware`**: Validates Host header in production mode
  - Throws `InternalErrorException` if `App.fullBaseUrl` is not configured
  - Throws `BadRequestException` if Host header doesn't match configured URL
  - Skips validation in debug mode (development)
- **Updated `Application.php`**: Adds middleware after `ErrorHandlerMiddleware`
- **Simplified `bootstrap.php`**: Removes the exception throwing, keeps development fallback

### Benefits

- Errors are properly rendered by `ErrorHandlerMiddleware`
- Logging works correctly
- CLI commands are not affected
- Additionally validates that incoming Host header matches the configured URL

Fixes concerns from #1071.

### DIFF
--- a/src/Middleware/HostHeaderMiddleware.php
+++ b/src/Middleware/HostHeaderMiddleware.php
@@ -1,0 +1,58 @@
+<?php
+declare(strict_types=1);
+
+namespace App\Middleware;
+
+use Cake\Core\Configure;
+use Cake\Http\Exception\BadRequestException;
+use Cake\Http\Exception\InternalErrorException;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * Middleware to validate Host header and prevent Host Header Injection attacks.
+ *
+ * In production, this middleware ensures that App.fullBaseUrl is configured
+ * and validates incoming Host headers against it. This prevents attackers
+ * from manipulating password reset links and other security-critical URLs.
+ *
+ * @see https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/07-Input_Validation_Testing/17-Testing_for_Host_Header_Injection
+ */
+class HostHeaderMiddleware implements MiddlewareInterface
+{
+    /**
+     * Process the request and validate the Host header.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Server\RequestHandlerInterface $handler The request handler.
+     * @return \Psr\Http\Message\ResponseInterface A response.
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        if (Configure::read('debug')) {
+            return $handler->handle($request);
+        }
+
+        $fullBaseUrl = Configure::read('App.fullBaseUrl');
+        if (!$fullBaseUrl) {
+            throw new InternalErrorException(
+                'SECURITY: App.fullBaseUrl is not configured. ' .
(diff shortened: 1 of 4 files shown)


dereuromark:
The test suite now passes (both PHP 8.2 lowest and PHP 8.5 highest).

The PHPStan failure is a pre-existing issue unrelated to this PR - it's about `AppView` not specifying generic types for `Cake\View\View`. This would need to be addressed separately.

ADmad:
Given that pretty much every publicly accessible website would be using HTTPS these days, we could use the `SSL_TLS_SNI` env var to determine the URL host instead of forcing users to set the full base URL. `SSL_TLS_SNI` can't be spoofed and is unaffected by user provided `Host` header. Depending on the web server config, some web servers themselves throw an error is the `Host` doesn't match the SNI.

dereuromark:
Should we on top still add the middleware to core for people to opt in additionally?

ADmad:
Given that the middleware checks for an app config, I think it can stay in the app repo.

dereuromark:
OK, feel free to adjust this PR as necessary/desired.

src/Middleware/HostHeaderMiddleware.php
```diff
@@ -0,0 +1,58 @@
...
+ * Middleware to validate Host header and prevent Host Header Injection attacks.
+ *
+ * In production, this middleware ensures that App.fullBaseUrl is configured
+ * and validates incoming Host headers against it. This prevents attackers
+ * from manipulating password reset links and other security-critical URLs.
+ *
+ * @see https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/07-Input_Validation_Testing/17-Testing_for_Host_Header_Injection
+ */
+class HostHeaderMiddleware implements MiddlewareInterface
+{
```
markstory, line 24: If this proves useful and problem free, I don't see why we couldn't integrate this into the BaseApplication logic.

LordSimal: where would you put that in the BaseApplication?

LordSimal:
This still leaves the question open how users with a multi domain setup would properly/nicely use this feature.

rochamarcelo:
Shouldn't this be responsability of the server? What if you have domain.com, api.domain.com and blog.domain.com poiting to the same app?
I'm not sure this middleware being loaded by default is ideal. It is strange to have 'real' middleware inside the app, in case of changes the user won't get a new version easily.

LordSimal:
> Shouldn't this be responsability of the server?

Not every webserver setup is built around catching/preventing Host Header Injection, thererfore it's good pratice and [seen in other frameworks](https://www.drupal.org/docs/getting-started/installing-drupal/trusted-host-settings) that the application itself defines, which domains/hosts are actually allowed for the current application.

I'd agree, that in the end this should be part of the core and should allow a user to set an array (like Drupal does it) of possible regex patterns to make it as flexible as possible.

src/Middleware/HostHeaderMiddleware.php
markstory: I was thinking it could go  in `BaseApplication::handle()` before `controllerFactory->invoke()`. However, thinking about it more, we give developers more flexibility with middleware as they can opt in/out or replace the behavior as required.

LordSimal:
lets go with this for now. It can be moved to the core for the 6.x app template later on. At least lets get a working app template out there which shows a nice error message in production mode.

PR #1096, opened 2026-07-03
Add misssing Error class import

ADmad:


dereuromark:
Wait, but since this is a non namespace file ( a template), I think its generally not needed, as root is auto-assumed.
PHPStorm for me displays it as superfluous and removable.

ADmad:
Yup, my bad, it's not needed. I'll revert.

